PR #617, opened 2022-08-15
fix: navbar items issue in support portal

kamaljohnson:
#### Before
The top navbar items did not consider parent/children layout, it showed all the items in navbar table

#### After
The child items in the navbar table are shown in the dropdown list of the parent navbar item


harshit-30:
please add screenshots to give context.

kamaljohnson:
@harshit-30 added the screenshots

PR #712, opened 2022-09-07
fix: misc bug fixes and typo

shadrak98:
#TODO

### DIFF
--- a/frappedesk/api/ticket.py
+++ b/frappedesk/api/ticket.py
@@ -66,7 +66,7 @@ def create_new(values, template='Default', attachments=[], via_customer_portal=F
 
 	return ticket_doc
 
-@frappe.whitelist(allow_guest=True)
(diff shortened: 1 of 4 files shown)


frappedesk/api/ticket.py
```diff
@@ -139,7 +142,7 @@ def assign_ticket_type(ticket_id, type):
 
 		return ticket_doc
 
-@frappe.whitelist(allow_guest=True)
+@frappe.whitelist()
 def assign_ticket_status(ticket_id, status):
```
kamaljohnson, line 146: There will be an issue here
if the ticket is created via email and then the user goes to the customer portal and tries to close the ticket, it will show an error because, writing permission for a ticket is only given to the creator of the ticket, but since the ticket was created via email the creator is set as admin.

PR #763, opened 2022-11-03
feat: add priority value for assignment rules created for agent groups

kamaljohnson:


harshit-30:
What feature is this? where is the description? @kamaljohnson

kamaljohnson:
@harshit-30 I have updated the PR title and description.

it was actually a fix.

PR #852, opened 2022-12-01
feat(wip): reply with kb article

esafwan:
Adds option to reply to tickets using Kb articles similar to that of canned responses.


harshit-30:
@fadilsid can we move forward with this?

PR #940, opened 2023-01-11
chore: update version v0.5.14 -> v0.5.15

kamaljohnson:


harshit-30:
why there is a disparity between the version of the application and in GitHub releases.

PR #950, opened 2023-01-18
fix: update sla based on condition

fadilsid:
Resolves sla not updating based on condition in tickets.


fadilsid:
@netchampfaris please review.

netchampfaris (commented):
Now that I think about it, the `or_filters` was there in case no SLA was found and it should keep the SLA that was applied. So, the logic should be "find an applicable SLA, if it is not found then return the SLA that is already applied"

harshit-30:
@fadilsid did you review what @netchampfaris has mentioned here?

fadilsid:
> @fadilsid did you review what @netchampfaris has mentioned here?

yes, if not sla is found for a ticket it applies the previously applied sla of the ticket.

PR #971, opened 2023-02-01
fix(ui): tickets: id column width

ssiyad:
### Problem
Id column width in tickets view is set as a constant value. (2 currently). While this works for out of the box setups, it might not work for dynamic formats (eg: expression).

### Investigation
Interestingly, this used to be dynamic
```
width: manager.totalCount > 999 ? '2' : '1'
```
This behavior was changed in 7329f863bdfb553c283627078faa3269ea130ed6. While this is not perfect, it could have worked for at-least some cases.

### Solution
I introduced another logic to calculate width, depending on the widest (longest string) id from the table. It is safe to assume that all ids will be of same length, and this logic will work for most cases. There could be room for improvement though, as more cases pop up.

ref: https://github.com/frappe/helpdesk/issues/958

### DIFF
--- a/desk/src/components/global/ListManager.vue
+++ b/desk/src/components/global/ListManager.vue
@@ -57,6 +57,45 @@ export default {
 			return listResource.list.loading
 		})
 
+		const idWidth = computed(() => {
+			// Default width for id column in cases where calculation is impossible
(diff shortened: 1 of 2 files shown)


ssiyad:
Succeeded by #983

PR #1048, opened 2023-03-12
feat(ui): update toast ui (fetched from frappe-ui)

ssiyad:
#### Customised to include

- [x] Buttons
- [ ] Form


ssiyad:
### Before
image

### After
image

PR #1155, opened 2023-05-09
fix: Ticket Activity doesn't exist.

kurogeek:
When clicking on `TICKET HISTORY` on a ticket page, there is an error saying, a table Ticket Activity doesn't exist.

This PR fix this issue. By changing `tabTicket Activity` to `tabHD Ticket Activity`.


ssiyad:
LGTM

PR #1187, opened 2023-05-17
fix: empty responses should not be allowed.

kurogeek:
Responses (Replies/Comments) should not be empty.

I don't know if this is the right way to do it. If there's another way to do this, please advice.


ssiyad:
This looks good. However, there is a `validate` function in frappe-ui intended for this. Check https://frappeui.com/resources/resource.html

kurogeek:
Got it, I'll fix it.

kurogeek:
@ssiyad Just fixed it, please review and let me know if there is any thing wrong.

Thank you for your guide BTW.

ssiyad:
@kurogeek Hey, are you working on this? Can you check the review and make necessary changes?

kurogeek:
Yes I am, let me try to fix it.

kurogeek:
Seems like the error is about a commit title violate commitlints. I'm going to close this one and recreate a new PR. So that I can properly add commit message.

PR #1198, opened 2023-05-18
fix(ux): ticket page: can not back go back if no history

ssiyad:
There is nothing to go back to if there is no history. This PR adds a check if there is a way to go back. If not, it will go to the default page


ssiyad:
closes https://github.com/frappe/helpdesk/issues/1196

PR #1225, opened 2023-05-25
fix: empty responses should not be allowed.

kurogeek:
Responses (Replies/Comments) should not be empty.

### DIFF
--- a/desk/src/pages/desk/ticket/data.ts
+++ b/desk/src/pages/desk/ticket/data.ts
@@ -3,6 +3,7 @@ import { defineStore } from "pinia";
 import { createDocumentResource } from "frappe-ui";
 import { socket } from "@/socket";
 import { createToast } from "@/utils/toasts";
+import { isEmpty } from "@/utils/editor";
 
 export const useTicketStore = defineStore("ticket", () => {
 	const sidebar = reactive({
@@ -47,6 +48,17 @@ export const useTicketStore = defineStore("ticket", () => {
 				debounce: 500,
 				onSuccess() {
 					clean();
+				},
+				validate() {
+					if (isEmpty(editor.content)) {
+						createToast({
+							title: "Message is empty",
+							icon: "x",
+							iconClasses: "text-red-600",
+						});
+
+						return "Message is empty";
+					}
 				},
 				onError(error) {
 					createToast({
--- a/desk/src/pages/desk/ticket/data.ts
+++ b/desk/src/pages/desk/ticket/data.ts
@@ -51,12 +51,6 @@ export const useTicketStore = defineStore("ticket", () => {
 				},
 				validate() {
 					if (isEmpty(editor.content)) {
-						createToast({
-							title: "Message is empty",
-							icon: "x",
-							iconClasses: "text-red-600",
-						});
-
 						return "Message is empty";
 					}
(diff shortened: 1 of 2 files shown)


kurogeek:
@ssiyad How can I run the check again? I think the error doesn't come from the code. Instead, it came from something else.

ssiyad:
@kurogeek `validate()` does not have to create a toast. It can return an error message which can be handled in `onError`. Can you make the change so that only `onError` create toasts

kurogeek:
@ssiyad got it, done.

desk/src/pages/desk/ticket/data.ts
```diff
@@ -48,6 +49,11 @@ export const useTicketStore = defineStore("ticket", () => {
 				onSuccess() {
 					clean();
 				},
+				validate() {
+					if (isEmpty(editor.content)) {
```
ssiyad: Use `params`

desk/src/pages/desk/ticket/editor/BottomSection.vue
```diff
@@ -140,6 +142,19 @@ const insertRes = createResource({
...
 	},
+	validate(params) {
+		if (isEmpty(params.doc.content)) {
+			return "Message is empty";
+		}
+	},
+	onError(error) {
+		createToast({
+			title: "Error sending reply",
+			text: error.messages?.join("\n"),
```
ssiyad, line 152: Error from `validate()` will be available as an `Error` object. You can get the message from this with `error.message`

ssiyad (changes requested):
There was a recent addition to editor. A `tiptap` object is added to `editor` which expose all utilities from it. Hence, you can use `editor.tiptap?.isEmpty` instead of `isEmpty` from helpdesk utils

kurogeek:
There is an error `params` is defined but never used. 
Is that okay?

ssiyad:
> There is an error params is defined but never used

You can remove it if not used

desk/src/pages/desk/ticket/editor/BottomSection.vue
```diff
@@ -140,6 +142,19 @@ const insertRes = createResource({
 	onSuccess() {
 		clean();
 	},
+	validate(params) {
+		if (isEmpty(params.doc.content)) {
```
ssiyad: We can use `isEmpty` from `tiptap` here too

ssiyad (changes requested):

desk/src/pages/desk/ticket/editor/BottomSection.vue
kurogeek: I tried, but it didn't work, I was still able to add an empty comment.

kurogeek: I used this though `params.doc.tiptap?.isEmpty`

kurogeek: or should I just use `editor` object instead of `params`?

kurogeek: I just went on using `editor` object anyway.

ssiyad (approved):
Looks good. Merging

PR #1319, opened 2023-07-19
feat(ux):username added in ticket history

Shivam327:
Username added to ticket history tab,
As shown in the Demo image of the product, now other users(agents) will be able to see which user has changed the state.
and username will be reflected in the ticket history timeline.


ssiyad:
I don't see how this fix the issue. The change you did doesn't do anything useful. This is the method which does the job

image

Issue #1269, opened 2023-06-14
UX: Reply/Comment should NOT be same button


ankush:
fairly easy to shoot yourself in foot by sending internal comment to end user

![image](https://github.com/frappe/helpdesk/assets/9079960/c151c8f5-4106-4fb4-bb0c-525ed3f841df)

Shivam327:
This issue is Solved.

PR #1320, opened 2023-07-19
fix(ux): separated reply and comment button

Shivam327:
As mentioned in issue #1269 
The issue has been tried to solve by creating a sperate button for mail reply and comment buttons.
Comment button is grey color (Tried to keep it consitent with other buttons).
Reply Button is in red color to refelect its presnce.
Suggestion: Another name should be given to reply button. as term may be confusing sometimes.


ssiyad:
@Shivam327 Can you format files? You can see Reviewdog complains in `Files Changed` tab

Shivam327:
@ssiyad Code formatted please check.

PR #1556, opened 2023-09-12
Update hd_ticket.py

ovresko:
frappe.get_value does not guarantee returned  fields order from database, we have situations where values are reversed between enable_restrictions and ignore_restrictions (one's take the other field value)

`value = get_list(  <============== IN FRAPPE THIS DOES NOT GUARANTEE FIELDS ORDER
			doctype,
			filters=filters,
			fields=fields,
			debug=debug,
			limit_page_length=1,
			parent=parent,
			as_dict=as_dict,
		) 

	if as_dict:
		return value[0] if value else {}

	if not value:
		return

	return value[0] if len(fields) > 1 else value[0][0]
`

fixing by reading the result

### DIFF
--- a/helpdesk/helpdesk/doctype/hd_ticket/hd_ticket.py
+++ b/helpdesk/helpdesk/doctype/hd_ticket/hd_ticket.py
@@ -74,7 +74,7 @@ class HDTicket(Document):
 		)
 		query = query.where(Criterion.any(conditions))
 
-		restricttions = frappe.get_value(
+		restrictions = frappe.get_value(
 			doctype="HD Settings",
 			fieldname=[
 				"restrict_tickets_by_agent_group",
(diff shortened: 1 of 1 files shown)


helpdesk/helpdesk/doctype/hd_ticket/hd_ticket.py
```diff
@@ -74,15 +74,19 @@ def get_list_filters(query: Query):
 		)
 		query = query.where(Criterion.any(conditions))
 
-		enable_restrictions, ignore_restrictions = frappe.get_value(
+		restricttions = frappe.get_value(
```
kingraphaii: spelling :)

kingraphaii (changes requested):

ankush:
> frappe.get_value does not guarantee returned fields order from database, we have situations where values are reversed between enable_restrictions and ignore_restrictions (one's take the other field value)

This isn't possible. Can you share a reproducible example?

ovresko:
get_value internally uses get_list
reference: https://github.com/frappe/frappe/blob/7edb80bf5cb96700aaa14a2f6a7abe09afde4bbc/frappe/client.py#L129

get_list does not guarantee fields order since they are mapped to a dict based on doctype structure
any instance, set enable_restrictions to True, create 2 teams enter as one and create ticket for the other you'll see the ticket.
enable_restrictions taking ignore_restrictions's  value and ignore_restrictions  taking enable_restrictions's value

ankush:
@ovresko this bug should really be fixed in framework. I'll try it out. Thanks for explanation :+1:

RitvikSardana:
Not related to HD

PR #1560, opened 2023-09-15
feat: list_view: refactor

ssiyad:
- Include navigation by default
- Make use of `resource`
- Auto load next page
- Number of loaded/total entries in navbar


ssiyad:
Related - https://github.com/frappe/helpdesk/issues/1532

PR #1606, opened 2023-10-23
patch(ticket): first response from communication

ssiyad:
This used to be set from `Replied` status. The logic is then changed to fetch first response from communication instead

### DIFF
--- a/helpdesk/helpdesk/doctype/hd_ticket/patches/first_responded_on.py
+++ b/helpdesk/helpdesk/doctype/hd_ticket/patches/first_responded_on.py
@@ -1,0 +1,23 @@
+import frappe
+
+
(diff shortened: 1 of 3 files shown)


helpdesk/helpdesk/doctype/hd_ticket/patches/first_responded_on.py
```diff
@@ -0,0 +1,23 @@
+import frappe
+
+
+def execute():
+	for t in frappe.get_all("HD Ticket"):
+		try:
+			c = frappe.get_last_doc(
```
ankush: Why fetch entire doc? Just do `frappe.db.get_value`?

Patch will run much faster :smile:

PR #1612, opened 2023-10-25
refactor: comment

ssiyad:


michellealva:
@ssiyad does this also include editable comments?

PR #1617, opened 2023-11-04
Ticket visibility for agents who are also a customer or lead

cwerner1:
fix https://github.com/frappe/helpdesk/issues/1616
If the agent also exists as customer or lead, they were not able to see all Tickets.

### DIFF
--- a/helpdesk/helpdesk/doctype/hd_ticket/hd_ticket.py
+++ b/helpdesk/helpdesk/doctype/hd_ticket/hd_ticket.py
@@ -83,9 +83,12 @@ class HDTicket(Document):
 			if not is_agent()
 			else []
 		)
-		customer = get_customer(user)
-		for c in customer:
-			conditions.append(QBTicket.customer == c)
+		
+		if not is_agent():
+			customer = get_customer(user)
+			for c in customer:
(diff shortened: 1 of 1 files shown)


cwerner1:
My fix doesn't work completly:

Because the function is used in the customer Portal and in the desk portal. You would see all Tickets also in the Customer Portal, which is not intended.

ankush:
Looks like I also fixed this without knowing this PR existed: https://github.com/frappe/helpdesk/commit/fc32d7740768c822d94d43a4626772ca96b28be1

cwerner1:
@ankush Please only have a Look at this "problem"

ankush:
@cwerner1 is there ever a case where single user is both agent and customer? 

I don't get it otherwise. can you share it in separate issue?

cwerner1:
@ankush Yes, In my case the helpdesk users are also using their system to create Issues, which the IT Department has to solve.
So I have Customers who can create tickets on the Customer Portal or mail.
The Helpdesk Agents who can create Tickets on the Customer Portal, Mail or the helpdesk.


I think if #1628 is solved my Issue is also solved - as my problem is a related to this security problem.

PR #1624, opened 2023-11-09
fix: fetch only HD Customer contacts

rtdany10:
Currently, system doesn't check if the fetched contact belongs to an HD Customer or not.
Added a where condition to fetch only HD Customer.


rtdany10:
@ssiyad

rtdany10:
@ssiyad

rtdany10:
@ssiyad please check, it's a very simple fix and a required one

rtdany10:
@ssiyad @ankush

PR #1645, opened 2023-12-07
fix: change Resolve btn to 'Resolve and Close' on portal

shadrak98:
On Resolve and Close, change the status to close

[On Resolve and Close, change the status to close](https://github.com/frappe/helpdesk/assets/30501401/fbc1624b-2ab4-437a-9b29-f8d90fb8c227)

### DIFF
--- a/desk/src/pages/ticket/TicketCustomer.vue
+++ b/desk/src/pages/ticket/TicketCustomer.vue
@@ -15,7 +15,7 @@
         </Button>
         <Button
           v-if="showResolveButton"
-          label="Resolve"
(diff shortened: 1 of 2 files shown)


desk/src/pages/ticket/TicketCustomer.vue
```diff
@@ -15,7 +15,7 @@
         </Button>
         <Button
           v-if="showResolveButton"
-          label="Resolve"
+          label="Resolve and Close"
```
ankush:
```suggestion
          label="Close"
```

PR #1658, opened 2023-12-21
fix: show comment in email

shadrak98:
Fixes https://github.com/frappe/helpdesk/issues/1655


ankush:
@shadrak98 

https://github.com/frappe/helpdesk/issues/1714

PR #1660, opened 2023-12-21
Fix: Correct Doctype reference for "Ticket Type and Ticket Priority"

safvanhuzain:
@ssiyad The current code encounters an issue when selecting the filter "Ticket Type and Ticket Priority" within the "Based on" field. The problem arises due to an incorrect Doctype being referenced in lines 61 and 72.

image


sumitbhanushali:
fixed in https://github.com/frappe/helpdesk/pull/1703. 

Thanks for identifying the fix

PR #1663, opened 2023-12-21
Filter out communications coming from Notifications in on_communication_u…

ovresko:
ignore frappe notifications as they also trigger on_communication_update and cause ticket to become Open when Notification is triggered as it creates a Communication also.

The solution is not perfect, as the system could have a notification account name other then "Notifications" but wanted to point to the problem.


RitvikSardana:
Hey @ovresko 
In what scenario this happens ```c.sender_full_name=="Notifications"``` ?

RitvikSardana:
closing due to inactivity

PR #1692, opened 2024-02-12
chore: added temporary banner for offsite

sumitbhanushali:
Screenshot 2024-02-09 at 3 43 37 PM


shadrak98:
@sumitbhanushali Rather than adding it in code.. can we add it as a configuration from ticket settings or somewhere

Message can be configurable

sumitbhanushali:
> @sumitbhanushali Rather than adding it in code.. can we add it as a configuration from ticket settings or somewhere
> 
> Message can be configurable 

Yes I have reverted this change

Issue #1698, opened 2024-02-19
 Add support for retrieving the ticket type when creating a new ticket


zeel-codder:
Currently, System only ask for the `Subject` and `Description` when creating a new ticket. After the ticket is created, we can update the `Ticket Type`. 
We can give users the option to select the `Ticket Type` while they are creating the new ticket as well.

KanchanChauhan:
- This will help in filtering the tickets
- Assignment Rules will be more efficient

<img width="783" alt="Screenshot 2024-02-19 at 5 30 12 PM" src="https://github.com/frappe/helpdesk/assets/16913064/6081d2e8-d10a-474c-a1ef-1eb8f298e23c">

sumitbhanushali:
@KanchanChauhan yes the feature makes sense

KanchanChauhan:
@sumitbhanushali As suggested we have created a PR for this https://github.com/frappe/helpdesk/pull/1700

sumitbhanushali:
merged https://github.com/frappe/helpdesk/pull/1700

PR #1700, opened 2024-02-20
feat: add support for retrieving the ticket type when creating a new ticket

zeel-codder:
Issue: #1698

Updated the code to take the ticket type input.

### Output
Screenshot 2024-02-20 at 3 33 16 PM

Screenshot 2024-02-20 at 3 33 36 PM

Screenshot 2024-02-20 at 3 34 09 PM

### DIFF
--- a/desk/src/pages/ticket/TicketNew.vue
+++ b/desk/src/pages/ticket/TicketNew.vue
@@ -28,7 +28,6 @@
       <div class="relative w-full">
         <Autocomplete
           v-model="ticketType"
-          :body-classes="w - full"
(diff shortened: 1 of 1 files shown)


sumitbhanushali:
@zeel-codder reverting this since ticket type also impacts ticket priority and thus SLA

KanchanChauhan:
@sumitbhanushali Understood. How about we have a checkbox, which says _Allow ticket type selection to User_. If it is checked only then we show the ticket type while raising ticket.

netbridge360:
I too would like a checkbox!

PR #1705, opened 2024-02-24
fix: revert pr 1700

sumitbhanushali:
reverts https://github.com/frappe/helpdesk/pull/1700/ since ticket type also impacts ticket priority and thus SLA


netbridge360:
Can we please have to be an option for ticket type? A simple checkbox?

Issue #1712, opened 2024-03-02
Can't attach while replying to ticket


ankush:
Attachments don't seem to be working while replying to a ticket. 

Workaround: replying from desk works.

ankush:
This broke again @RitvikSardana :smile:

tdtce:
Hi, were you able to bypass this behavior? I have the same error, although the images are being sent correctly.

ankush:
this was fixed sometimes later, idk when/where but it works

Issue #1695, opened 2024-02-13
Attachment upload error while creating a new ticket


chavda-bhavik:
While creating a new ticket and uploading the attachment the upload_file API throws following error,
![image](https://github.com/frappe/helpdesk/assets/50201755/54d69077-3acd-4676-ab47-36bcc674a52d)

Here is the response,
```
{
    "exc_type": "ValidationError",
    "exception": "frappe.exceptions.ValidationError: Attached To Name must be a string or an integer",
[...]

PR #1715, opened 2024-03-06
fix: Attach attachments after the ticket is saved

niraj2477:
### Issue:
When we upload an attachment in ticket `doctype` as parameter is passed, which will breaks the file upload
https://github.com/frappe/frappe/blob/develop/frappe/core/doctype/file/file.py#L140

### Solution:
While uploading attachments, we do not need to pass the `doctype` parameter, it will be automatically handled by `create_communication_via_contact` method

issue ref: #1695 ,#1712


niraj2477:
@sumitbhanushali Can you have a look at this? As we are not able to upload the attachment because of this.

PR #1716, opened 2024-03-11
Ticket Agent Redesign

sumitbhanushali:


asharamseervi:
Hello @sumitbhanushali Greetings, your contribution to the helpdesk is incredible.

Just casual question on checking your way-forward or roadmap on Ticket Agent Redesign stuff, why we've invested that much effort in redesigning, and suddenly discarded all the efforts you've made!

I would like to understand, what value this redesign will bring to the helpdesk and what we're missing on discarding this redesign.

sumitbhanushali:
@asharamseervi redesign is not discarded...i will raise another PR for it which will be merged in develop first

PR #1727, opened 2024-03-19
feat : Added Email Signature 

Jishnusuni:
This enhancement enabling user to include an email signature when responding to tickets.Users can  add their email signature in  "Email Signature" field in "User" doctype.

image

whenever a user sends a response to a ticket, the email signature will be automatically appended with the message.

image

image


federicocalvo:
@Jishnusuni hi! this allows the incorporation of images?

PR #1768, opened 2024-04-30
feat: Implemented email sending for new ticket creation

Jishnusuni:
This enhancement enables users to send an email response when creating a new ticket. Within the Helpdesk UI, users can now select the new ticket type as either a 'comment' or 'response'. If the user selects the response type, they are able to send an email response. I have added the fields TO, CC, and BCC to facilitate email sending.

image

image


sumitbhanushali:
1) comments are meant for internal communication only and should not be customer facing.
2) why would a customer need to, cc, and bcc fields when creating a ticket?

PR #1854, opened 2024-07-10
customer listing in ticket activity

NihalRoshanCK:
customer listing fixed at the ticket activity right sidebar
helpdesk_bug

### DIFF
--- a/desk/src/components/ticket/TicketAgentFields.vue
+++ b/desk/src/components/ticket/TicketAgentFields.vue
@@ -44,6 +44,7 @@ import { Autocomplete } from "@/components";
 import { useTeamStore } from "@/stores/team";
 import { useTicketPriorityStore } from "@/stores/ticketPriority";
 import { useTicketTypeStore } from "@/stores/ticketType";
+import { useCustomerStore } from "@/stores/customer";
 import UniInput2 from "@/components/UniInput2.vue";
 
 const emit = defineEmits(["update"]);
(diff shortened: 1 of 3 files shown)


Nihalroshan55:
when we create a new ticket from helpdesk we can't add customer reference after creating that we have a view that shows all the ticket details in its sidebar there are details including customer and all if there is no customer (in case we create it from helpdesk ) 
from there customer listing is not working

rmehta:
can you send a clear video?

Nihalroshan55:
https://github.com/user-attachments/assets/cf27f31c-677e-4699-a403-35236426cec1

sumitbhanushali:
@Nihalroshan55 in your change, you are fetching data from Customer in agentstore? was this by mistake?

Also, Customers are fetched based on Contact -> Customer mapping. If a contact doesn't have customer mapped, It shows blank. This thing needs better UX but no change in functionality afaik

Nihalroshan55:
yes, it's done by mistake changed it back @sumitbhanushali

sumitbhanushali:
@Nihalroshan55 checkout `get_customer` api at helpdesk/utils.py.

basically, it fetches customer from given Contact. This contact must have mapping in Contact doctype in links child table

Nihalroshan55:
as in our use case,  we received a call for a product bug,  which we are trying to manually create a ticket on behalf of customer. Here we would like to create a customer manually or sync from our customer list. (this is a non frappe external use case)

RitvikSardana:
fixed via: https://github.com/frappe/helpdesk/pull/2063/commits/f82fdaa03cd5954cf66c2a15357c23f00a18fd10

PR #1860, opened 2024-07-13
fix: header breadcrumb styling

RitvikSardana:
**Issue:**
Screenshot 2024-07-13 at 12 36 26 PM

1. The breadcrumb UI looks cheap (it has a button in the breadcrumb) and has an on click functionality to update the "subject" of the ticket, which does not make any sense.

2. When the subject of the ticket is too long there is an overflow in screen.


sumitbhanushali:
https://github.com/frappe/helpdesk/issues/915 
https://github.com/frappe/helpdesk/issues/1523

editable subject is needed

PR #1918, opened 2024-08-14
feat(TicketsAgent): export tickets list in excel,csv format

sumitbhanushali:


RitvikSardana:
@sumitbhanushali please add some description and screenshots

RitvikSardana:
regardless of the rows selected the export is for all the records, so the checkbox does not have any meaning.

It exports all the tickets out of the box. Even if we uncheck the Checkbox. @sumitbhanushali

sumitbhanushali:
> regardless of the rows selected the export is for all the records, so the checkbox does not have any meaning.
> 
> It exports all the tickets out of the box. Even if we uncheck the Checkbox. @sumitbhanushali 
> 
> 
> 

It doesn't work with check boxes currently...it exports what is on current page or all...

That is what checkbox is used for

RitvikSardana:
It it not intuitive enough what rows will be exported, it should be done for the selected rows.

PR #1924, opened 2024-08-14
fix(HelpdeskSearch): Improve search of knowledge base articles

balamurali27:
## Before
image
## After
image

# To be done in UI

~- [ ] Fix ticket id search~ Not doing this as it's too out of the way. Filter already exists for this anyway
- [x] Show headings instead of subject
- [x] Show description in search results. Render mark tags
- [x] Put # in url instead of %23 
- [x] Replace the sql search in /my-tickets with HelpdeskSearch
- [x] Add permalinks to headings so urls will scroll to relevant section

### DIFF
--- a/helpdesk/api/article.py
+++ b/helpdesk/api/article.py
@@ -6,7 +6,6 @@ from helpdesk.search import search as hd_search
 @frappe.whitelist()
 def search(query: str):
 	out = hd_search(query, only_articles=True)
-	breakpoint()
 	if not out:
 		return []
 	return out[0].get("items", [])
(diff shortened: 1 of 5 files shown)


helpdesk/api/article.py
```diff
@@ -1,30 +1,12 @@
...
-	if len(query) < min_charecters:
-		return []
 
-	queries = query.split(" ")
-	queries = [query for query in queries if len(query) >= min_charecters]
-	if len(queries) == 0:
+@frappe.whitelist()
+def search(query: str):
+	out = hd_search(query, only_articles=True)
+	breakpoint()
```
RitvikSardana: please remove breakpoint

PR #1949, opened 2024-08-29
Fix: Reset Form Fields After Creating a New Category in Knowledge Base from Agent Portal

UmakanthKaspa:
**Summary**

This PR addresses an issue in the Knowledge Base category creation dialog where form fields (title, description, and icon) were not being reset after successfully creating a new category.

**Changes**

- Implemented a reset for form fields only after a new category is successfully created.
- No changes were made to the behavior when the dialog is closed without creating a category; the form data remains intact, allowing users to continue their input if the dialog is reopened.

**Testing**

- Verified that the form fields are reset only after successful creation of a category.
- Confirmed that form data persists if the dialog is closed without creating a category.

**Impact**

The form fields in the Knowledge Base category creation dialog will now reset only after a successful category creation. If the dialog is closed without creating a category, the form fields will retain their values, allowing users to continue with their input as usual.

### DIFF
--- a/desk/src/pages/knowledge-base/KnowledgeBaseCategoryNew.vue
+++ b/desk/src/pages/knowledge-base/KnowledgeBaseCategoryNew.vue
@@ -48,9 +48,9 @@ const newCategoryDescription = ref("");
 const newCategoryIcon = ref("");
 
 const resetForm = () => {
-  newCategoryName.value = '';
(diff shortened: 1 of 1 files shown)


UmakanthKaspa:
@RitvikSardana Thanks for the feedback! I'll make sure to include before and after screenshots in future PRs.

PR #1959, opened 2024-09-02
Fix: Filter Out Inactive Agents in Assign To Dialog

UmakanthKaspa:
### Summary
This PR fixes a bug where inactive agents were showing up in the "Assign To" dialog in the Agent Portal. 

### Changes Made
- Updated the dialog to filter out inactive agents, so only active ones appear in the list.

### Screenshots
image

**Before:**
image
*Inactive agents are shown in the list.*

**After:**
image
*Only active agents are shown.*

### DIFF
--- a/desk/src/components/AssignmentModal.vue
+++ b/desk/src/components/AssignmentModal.vue
@@ -9,9 +9,9 @@
     <template #body-content>
       <SearchComplete
         class="form-control"
-        value="1"
-        valueField="is_active"
+        value=""
         doctype="HD Agent"
+        :customFilters="customFilters"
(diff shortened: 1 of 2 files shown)


RitvikSardana:
**Provided Solution to filter**
image
This is the query which goes when you use the prop "value-field".


**Expected**
image

The correct way is to use the prop "search-field"  which filters the data in the query.

Also while using props if the prop name is "searchField" while passing it as a prop you should sent it like "search-field".

UmakanthKaspa:
image

@RitvikSardana please check  if this is okay?

PR #1996, opened 2024-09-25
fix(PWA): change PWA assets padding

RitvikSardana:
Update the assets for PWA. 
Handled padding for different screen sizes


mohsinalimat:
image

PR #2005, opened 2024-09-30
fix: Allow users to reorder fields in the `HD Ticket Template` for customizable display order

DaizyModi:
**Before**
When additional fields were added or reordered in the HD Ticket Template, they would not appear in the same order on the Ticket Portal.

**After**
Fields are now displayed on the Ticket Portal in the exact order set by the user in the HD Ticket Template.

Screenshot from 2024-09-30 12-45-29


DaizyModi:
> yarn.lock file should not be deleted.
> 
> create a new PR for fixing the issue

Fixed https://github.com/frappe/helpdesk/pull/2021

PR #2019, opened 2024-10-21
feat: settings to allow to create tickets without permission

RitvikSardana:
Introducing a new HD Setting  called "Allow anyone to create tickets".
If enabled This setting allows anyone to create tickets on your support portal. Useful in cases where you wants users to create tickets via webforms or some external APIs without authentication.

By default the value is 0.


jiguantong:
Is there any document about how to call this through API? I have tried to
```
post https://helix-support.vercel.app/api/resource/hd_ticket
```
but got   `Error: No module named 'frappe.core.doctype.hd_ticket'`

PR #2021, opened 2024-10-23
fix: Allow users to reorder fields in the `HD Ticket Template` for customizable display order

DaizyModi:
**Before**
When additional fields were added or reordered in the HD Ticket Template, they would not appear in the same order on the Ticket Portal.

**After**
Fields are now displayed on the Ticket Portal in the exact order set by the user in the HD Ticket Template.

Screenshot from 2024-09-30 12-45-29

### DIFF
--- a/helpdesk/helpdesk/doctype/hd_ticket_template/api.py
+++ b/helpdesk/helpdesk/doctype/hd_ticket_template/api.py
@@ -27,7 +27,7 @@ def get_one(name: str):
     fields.extend(get_fields(name, "Custom Field"))
     return {
         "about": about,
-        "fields": fields,
+        "fields": sorted(fields, key=lambda x: x["idx"]),
     }
 
 


RitvikSardana:
Better way was to use orderby while getting fields, which has already been done. (https://github.com/frappe/helpdesk/blob/develop/helpdesk/helpdesk/doctype/hd_ticket_template/api.py#L63)

Thanks for the PR.

DaizyModi:
> Better way was to use orderby while getting fields, which has already been done. (https://github.com/frappe/helpdesk/blob/develop/helpdesk/helpdesk/doctype/hd_ticket_template/api.py#L63)
> 
> Thanks for the PR.

Thank you for considering my PR! 

While `orderby` can define the sequence of fields, it doesn't achieve the desired outcome in this case. This is because the `get_one` function first appends DocFields to the list, followed by custom fields. As a result, the `orderby` clause in the `get_fields` function has no meaningful impact on the final order.

PR #2039, opened 2024-11-08
fix: add confimation dialog and use it to handle discard ticket

joshuarestivo:
Previously, the Discard button in the New Ticket view only cleared the contents of the editor. This was non-intuitive and not terribly useful. This PR creates a confirmation dialog component that is called by the Discard button. Selecting 'Yes' on the dialog routes the user back to the appropriate Ticket list view (Agent or Customer).
image


joshuarestivo:
Happy to get rid of package-lock.json. Regarding closing a ticket from the customer portal, currently when you do that, a 'Rate this ticket' dialog appears. Do you want the confirmation instead of the ticket rating?

I've never been a fan of forced rating myself. I guess we could put a switch in Settings to select the desired close action (Rate Ticket or Confirm and Close). 

Let me know your thoughts.

RitvikSardana:
Hey joshua, I added the confirmation dialog along with this feature #2058

samuelcseto:
@RitvikSardana
I guess you forgot to include the change made by @joshuarestivo, currently, the discard button still has no dialog and does not navigate to the Tickets page.

PR #2045, opened 2024-11-19
feat : Add custom status to tickets

Sampadakambagi:
If extra statuses are added in the doctype, they show up in the frontend.
Screenshot 2024-11-19 120148


RitvikSardana:
This wont work, because the whole helpdesk is tightly packed with only 4 default Status.

PR #2047, opened 2024-11-20
fix: Maintain order of custom fields.

Sampadakambagi:
This fix maintains the order in which custom fields were added to the tickets and shows up in the side panel.


vkolgi:
@RitvikSardana This PR here is fixing the issue we have with ordering the custom fields. 
Will you be able to review it ? Looks like a simple fix and has solved our issue with presenting the fields in NewTicket in the order they are created in the HD Ticket Template.

PR #2091, opened 2024-12-16
chore: use PEP 660 python install method pyproject.toml

FHenry:
fix: https://github.com/frappe/helpdesk/issues/1991

Use editable installs for pyproject.toml based builds (wheel based)

https://peps.python.org/pep-0660/


RitvikSardana:
closed by https://github.com/frappe/helpdesk/pull/2205

PR #2106, opened 2024-12-19
fix: routing

RitvikSardana:
1. On customer portal we were calling api to get all_users. So basically anyone with access to Customer Portal would be able to see the user database along with images.
image

2. Because we get all_users before_route the page load becomes slow, to improve it we will get all_users after routing, to increase performance.


RitvikSardana:
Invalid routing, creating another PR

PR #2155, opened 2025-01-28
perf: using db.count instead of get_list to calculate total count

### LINKED ISSUE #2148: perf: don't fetch all rows just to count them
https://github.com/frappe/helpdesk/blob/d893917454d4644ee0716c70a430eb273e458efb/helpdesk/api/doc.py#L297

KerollesFathy:
Fixes #2148

### DIFF
--- a/helpdesk/api/doc.py
+++ b/helpdesk/api/doc.py
@@ -294,7 +294,7 @@ def get_list_data(
         "data": data,
         "columns": columns,
         "fields": fields if doctype == "HD Ticket" else [],
-        "total_count": len(frappe.get_list(doctype, filters=filters)),
+        "total_count": frappe.db.count(doctype, filters=filters),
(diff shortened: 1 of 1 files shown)


RitvikSardana:
This won't work, db.count does not apply permissions, so even on the customer portal side, total count of tickets will be shown and not the total tickets created by the customer.

I have fixed this using get_list :https://github.com/frappe/helpdesk/pull/2151/commits/e7869352b5684a1518c27a3123af23e2358c9aec

kakde-sandeep:
@RitvikSardana count function has the filters parameter, can we use this to filter the records based on the user. Below is the function signature.
count(self, dt, filters=None, debug=False, cache=False, distinct: bool = True)

RitvikSardana:
yes but then it missed if an customer altogether wants to see the company's ticket

Issue #2141, opened 2025-01-14
Enable Forwarding of Emails with Attachments in Frappe Helpdesk Conversations


fmarslan:
### Is your feature request related to a problem? Please describe.

Yes, this feature request addresses a common problem we face when handling customer communications. Currently, there is no straightforward way to forward an email conversation along with its attachments to an external email address. This becomes frustrating in scenarios where:

A support ticket involves external collaborators or vendors who need access to specific conversations and their attachments.
We need to share emails with attachments manually, which requires downloading the files and composing a new email, adding unnecessary steps and potential errors.
This limitation slows down our workflow and increases the time spent managing email communications outside the Helpdesk system. A built-in forwarding option would eliminate these challenges and improve efficiency.

### Describe the solution you'd like

I would like to see a "Forward" option integrated into the Frappe Helpdesk conversation view. This option should allow users to forward an email, including all its attachments, to an external email address directly from the Helpdesk interface.

The solution should work as follows:

Forward Button: Add a clearly visible "Forward" button in the email conversation interface.
Recipient Field: Upon clicking the "Forward" button, a prompt or modal should appear where users can input the recipient's email address.
Include Attachments: Ensure that all attachments from the original email are automatically included in the forwarded email.
Email History: The forwarded email should retain the original email's content (body and subject line) for context.
Audit Trail: Keep a record of the forwarded email in the Helpdesk ticket for transparency and tracking purposes.
This feature would streamline the process of sharing email communications with external parties and save significant time for our team.

### Describe the alternatives you've considered

Manual Forwarding via Email Client:

We considered manually downloading the email content and attachments from the Helpdesk system, composing a new email in an external email client (e.g., Outlook or Gmail), and attaching the downloaded files.
Drawbacks: This process is time-consuming, prone to human error, and disrupts workflow by requiring users to switch between platforms.
Sharing Helpdesk Access with External Parties:

Another alternative is granting external collaborators access to the Helpdesk system so they can view the ticket details directly.
Drawbacks: This poses security and privacy concerns, as external parties would gain access to more information than necessary. It also requires additional user management and training.
Exporting Email Threads as PDFs:

Exporting email threads as PDFs and manually attaching them to a new email was another potential workaround.
Drawbacks: This method is cumbersome, does not include original email attachments in their native format, and can lead to formatting issues or loss of context.
Third-Party Integration:

Using third-party integrations or middleware to forward emails was considered.
[...]

PR #2265, opened 2025-04-17
Enable Email Forwarding with Attachments in Conversations #2141

fmarslan:
This PR adds functionality to forward email conversations along with their original attachments directly from the Helpdesk interface. It streamlines communication with external collaborators and addresses the limitations highlighted in issue #2141.​

### DIFF
--- a/desk/src/components/CommunicationArea.vue
+++ b/desk/src/components/CommunicationArea.vue
@@ -131,6 +131,15 @@ function replyToEmail(data: object) {
   );
 }
 
+function forwardEmail(data: object) {
+  showEmailBox.value = true;
+  emailEditorRef.value.addToForward(
+    data.content,
+    data.subject,
+    data.attachments
+  );
+}
+
 const props = defineProps({
   doctype: {
     type: String,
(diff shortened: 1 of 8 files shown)


RitvikSardana:
Please put a video of this feature

fmarslan:
helpdesk-forward-ticket

fmarslan:
Using CC and BCC in the reply is helpful, but forwarding has a different purpose. For example, when we forward an email, we usually want to include attachments and keep the original content without indentation. In replies, attachments are not included automatically and the message gets indented.

So instead of adding these features into the reply function, it makes more sense to have a separate forward feature to handle this properly.

fmarslan:
Hi, Just a gentle reminder regarding this PR. It's been open for a while, and I wanted to check if there's anything else needed from my side to move it forward.

RitvikSardana:
Unwanted feature as of now

michal-bee:
@RitvikSardana This is an important feature. Without it, agents are often forced to use a separate email client just to forward helpdesk emails—especially when attachments need to be included, which is a very common scenario.

PR #2285, opened 2025-04-29
fix: multiple filters for the same field

the-bokya:
https://github.com/user-attachments/assets/1ac3de6d-59ce-43c9-b3a4-e9e7708d27aa

https://github.com/user-attachments/assets/3130882d-ac6c-4677-9ee8-700fa31ae35e


helpdesk/api/doc.py
```diff
@@ -23,11 +23,10 @@ def get_list_data(
     is_default=False,
 ):
     is_custom = False
-
     rows = frappe.parse_json(rows or "[]")
     columns = frappe.parse_json(columns or "[]")
-    filters = frappe.parse_json(filters or "[]")
 
+    print(filters)
```
RitvikSardana: remove this

desk/src/components/view-controls/Filter.vue
```diff
@@ -191,16 +190,18 @@ const listViewActions = inject("listViewActions");
...
-  if (!allFilters || !filterableFields.data) return new Set();
-
+  if (!allFilters || !filterableFields.data) return new Array();
   return convertFilters(filterableFields.data, allFilters);
 });
 
 function convertFilters(data, allFilters) {
   let f = [];
-  for (let [key, value] of Object.entries(allFilters)) {
+  for (let i of allFilters) {
```
the-bokya, line 201: this assumes each item to be an array
but in our case the first is object.


so i[0], i[1] and all will not work hence the bug

RitvikSardana:
closing due to inactivity

PR #2300, opened 2025-05-18
fix: email threading

RitvikSardana:
**Issue:**
1. Emails are not threaded for the tickets that are created via "Mail", which causes issues in understanding the context
2. No acknowledgement email is sent to the tickets that are created via "Mail"
3. To pull the latest Emails, we have to go to the desk and click on the "Pull Emails" button to get them

**Solution:**
1. Emails are now threaded using "message id" & "in_reply_to" headers for the related emails, only for tickets created via email.
2. Acknowledgement email is sent
image

3. Pull Emails button is now present on the Frontend
Screenshot 2025-05-18 at 10 11 59 PM

### DIFF
--- a/helpdesk/helpdesk/doctype/hd_ticket/hd_ticket.py
+++ b/helpdesk/helpdesk/doctype/hd_ticket/hd_ticket.py
@@ -427,7 +427,7 @@ class HDTicket(Document):
     ):
         skip_email_workflow = self.skip_email_workflow()
         medium = "" if skip_email_workflow else "Email"
-        subject = f"Re:[## {self.name} ##] {self.subject}"
+        subject = f"Re: {self.subject}"
         sender = frappe.session.user
         recipients = to or self.raised_by
         sender_email = None if skip_email_workflow else self.sender_email()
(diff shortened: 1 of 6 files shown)


helpdesk/helpdesk/doctype/hd_ticket/hd_ticket.py
```diff
@@ -534,6 +522,8 @@ def reply_via_agent(
                 subject=subject,
                 template=template,
                 with_container=False,
+                in_reply_to=last_communication.name,
+                message_id=last_communication.message_id,
```
ankush, line 526: This doesn't seem correct. 

messaeg IDs ae unique, and in-reply-to is also a message id, not some name.

helpdesk/helpdesk/doctype/hd_ticket/hd_ticket.py
```diff
@@ -476,6 +460,9 @@ def reply_via_agent(
                 "subject": subject,
             }
         )
+        if last_communication and last_communication.message_id:
+            communication.in_reply_to = last_communication.name
+            communication.message_id = last_communication.message_id
```
ankush, line 465: same problem

PR #2301, opened 2025-05-20
fix(core): Fix email validation

iptelephony:
1. Use RFC-5322 compliant email address regular expression
2. Split email address list taking into account commas in display name

### DIFF
--- a/desk/src/components/EmailEditor.vue
+++ b/desk/src/components/EmailEditor.vue
@@ -257,6 +257,7 @@ function submitMail() {
     return;
   }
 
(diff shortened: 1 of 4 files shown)


desk/src/components/EmailEditor.vue
```diff
@@ -257,6 +257,7 @@ function submitMail() {
     return;
   }
 
+  console.log(toEmailsClone.value.join(","));
```
RitvikSardana, line 260: remove this please

yarn.lock
RitvikSardana, line 1: dont delete this file

PR #2324, opened 2025-06-04
fix: return agent name instead of agent doc while finding assigned agents

RitvikSardana:
We return the Agent doc instead of the agent name while getting the current assigned agents of the "agent_group".

This causes an error because

```
frappe.db.exists(
            "HD Team Member",
            {
                "parent": team,
                "user": agent,
            },
        )
```
this code takes `agent` name instead of agent doc, hence if agent doc is passed we get an SQL error operation


niraj2477:
@RitvikSardana  Can you take a look at this once? https://github.com/frappe/helpdesk/pull/2294

PR #2328, opened 2025-06-08
fix: custom statuses

RitvikSardana:


RitvikSardana:
Yes it means that, but this was a work around so wont be going ahead with this. 
Will make another PR with better solution

PR #2368, opened 2025-06-27
New Feature: customer can record screen video and attach it to ticket

abdo-host:
Add new button in ticket conversation to record video from share screen and save it in rickets attachments

image


RitvikSardana:
Please use conventional commits

+

Make a video of the feature, working on both agent and customer side

abdo-host:
This feature in the customer portal

RitvikSardana:
Should be available on both sides

Where we have merge tickets option you can add this feature on the agent side as well

PR #2386, opened 2025-07-08
docs: add missing migration and build steps to local setup

satriadhikara:
## Problem
The current local setup instructions in the README are missing critical steps that cause the helpdesk to show a blank page after installation. Users following the current steps will encounter:
- Blank page when accessing http://helpdesk.test:8000/helpdesk
- Missing frontend assets
- Database migrations not applied

## Solution
Added two missing steps to the local setup instructions:
- `bench --site helpdesk.test migrate` - Applies database migrations
- `bench build --app helpdesk` - Builds frontend assets

## Testing
I experienced this issue firsthand and confirmed these steps resolve the blank page problem.

## Checklist
- [x] Tested the solution personally
- [x] Updated documentation with clear, actionable steps
- [x] Follows conventional commit format
- [x] No breaking changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,8 @@ To setup the repository locally follow the steps mentioned below:
 1. Map your site to localhost with the command `bench --site helpdesk.test add-to-hosts`
 1. Get the Helpdesk app. Run `bench get-app https://github.com/frappe/helpdesk`
 1. Run `bench --site helpdesk.test install-app helpdesk`.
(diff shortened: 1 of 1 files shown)


RitvikSardana:
I don't think migrate step is needed, can you confirm?

and also if it is the case, migrate should come after bench build

satriadhikara:
Yeah you right the migrate step is not needed sorry, i'll change that

but the build is still necessary.

PR #2392, opened 2025-07-10
fix(security): private media

RitvikSardana:
1. Before sending mail to the customer, add 'embed' attribute in the img/video element to handle inline content, 'embed' attribute is important because framework replaces it with `` this in turn is displayed as an image in the mail sent to the customer.

2. All media are now private and are attached to the relevant document

2. Add  `uploadFunction` prop in TextEditor to attach the file to the document, basically attaching video/image to the ticket
3. All scenarios are covered to ensure privacy for user's media 
- New ticket raised via portal
- Ticket raised via email
- Reply from agent
- Reply from customer

5. Delete removed attachments from the server
6. Preview images before sending
7. Update frappe-ui to 0.1.168 and update yarn.lock file

### DIFF
--- a/helpdesk/helpdesk/doctype/hd_ticket/hd_ticket.py
+++ b/helpdesk/helpdesk/doctype/hd_ticket/hd_ticket.py
@@ -5,6 +5,7 @@ from functools import lru_cache
 from typing import List
 
 import frappe
+from bs4 import BeautifulSoup
 from frappe import _
 from frappe.core.page.permission_manager.permission_manager import remove
 from frappe.desk.form.assign_to import add as assign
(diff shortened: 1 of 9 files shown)


helpdesk/helpdesk/doctype/hd_ticket/hd_ticket.py
```diff
@@ -631,6 +635,24 @@ def create_communication_via_contact(
...
+            if tag.has_attr("src"):
+                src = tag["src"]
+                # this is an inline media, we need to add it to the attachments
+                files.append(src)
+
+        for f in files:
+            last_doc = frappe.get_last_doc(
+                "File", {"file_url": f, "attached_to_doctype": ["!=", "Null"]}
+            )
+            if last_doc:
```
ankush: This will always be present. If matching doc is not found then `get_last_doc` raises an error.

RitvikSardana: Using 'get_last_doc' for the edge case when the file name uploaded by two different users is same.

Any other suggestion?

PR #2398, opened 2025-07-14
feat: add assignment rule

pratikb64:
This PR adds functionality to add and manage Assignment rule from the frontend

### Assignment Rule List View
image

### Assignment Rule View
image

image


desk/src/components/Settings/Agents.vue
```diff
@@ -136,6 +136,7 @@
     </div>
   </div>
   <AddNewAgentsDialog
+    title="Add Agents"
```
RitvikSardana: use translation syntax

desk/src/stores/assignmentRules.ts
```diff
@@ -0,0 +1,162 @@
...
+  {
+    day: "Friday",
+  },
+  {
+    day: "Saturday",
+  },
+  {
+    day: "Sunday",
+  },
+];
```
RitvikSardana: this format seems unnecessary

PR #2424, opened 2025-07-31
fix: sync standard and quick list filters

### LINKED ISSUE #2419: Filter isnt persistent
Steps:

- Add something using standard filter.... for example, I've entered "welcome" as subject in the standard filter.
- It gets applied. Now refresh. The "welcome" is deleted from standard filter. However the filter prevails.

Image

EltonLobo07:
Closes #2419

> We will have to come up with a better way to sync the filters (or not sync them at all) once we work on [this](https://github.com/frappe/helpdesk/issues/2256) issue.

There are 2 types of filters

filter-types

Until now, updating quick filters has automatically updated the standard filters, but this update was partial (see [this](https://github.com/frappe/helpdesk/issues/2419) issue). This PR fully syncs the quick filters with the standard filters whenever it makes sense.

https://github.com/user-attachments/assets/a804c464-36e5-470b-a62b-2eb005f47ce3

### DIFF
--- a/desk/src/components/ListViewBuilder.vue
+++ b/desk/src/components/ListViewBuilder.vue
@@ -521,7 +521,7 @@ const showViewControls = computed(() => {
     filterableFields.data &&
     sortableFields.data &&
     quickFilters.data &&
-    list.data?.data.length > 0
+    list.data?.data?.length > 0
   );
 });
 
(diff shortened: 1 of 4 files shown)


RitvikSardana:
Remove % from the UI @EltonLobo07

RitvikSardana:
Feels confusing for the end user to see the %, normal users dont care about SQL patterns

PR #2475, opened 2025-08-11
fix: framework code adjustments to disallow inviting disabled users

EltonLobo07:


michellealva:
@ankush will you be reviewing this?

PR #2483, opened 2025-08-13
feat: add telephony app integration & call logs page

pratikb64:
Add telephony app integration & call logs page

### Telephony settings
image

### Diler UI
image

### Call logs
image


desk/src/components/layouts/AppHeader.vue
```diff
@@ -1,7 +1,15 @@
 <template>
   <div class="flex border-b pr-5">
     <div id="app-header" class="flex-1 w-full"></div>
+    <div class="flex items-center justify-center">
+      <CallUI :userEmail="user" />
```
RitvikSardana, line 5: we dont need to send props here

u can easily get this value directly from the session store, 
please check CRM

this prop is not required

RitvikSardana (commented):
Pending

desk/src/components/ticket/TicketAgentContact.vue
```diff
@@ -11,14 +11,18 @@
...
+              <FeatherIcon
+                name="phone"
+                class="size-full p-1.5"
+                @click="callContact"
+              />
             </template>
           </Button>
         </Tooltip>
         <Tooltip :text="contact.email_id">
           <Button class="h-7 w-7">
```
RitvikSardana: The click handle should be on button not the icon, you have done a hack to solve the issue. Just sematically adding it to the button works.

pratikb64: Because of Tooltip click handler on button is getting called twice, that is why I have done this

PR #2553, opened 2025-08-29
fix: on mention comment skip email if skip_email_workflow is enabled

sumitbhanushali:
Issue: Comments where users are mentioned throw error that it needs email account configured even if skip_email_workflow is True
Fix: skip sending email when skip_email_workflow is checked


sumitbhanushali:
> why main-hotfix? why not develop?

I need this fix for version-15

RitvikSardana:
What about people who are using develop branch? Please think from global perspective, not just yours

sumitbhanushali:
This PR can't be used to also port to develop?

PR #2554, opened 2025-08-29
fix: on mention comment skip email if skip_email_workflow is enabled

sumitbhanushali:


helpdesk/helpdesk/doctype/hd_notification/hd_notification.py
```diff
@@ -50,6 +50,10 @@ def get_args(self):
 
     def after_insert(self):
         if self.notification_type == "Mention":
+            settings = frappe.get_single("HD Settings")
```
RitvikSardana: use "frappe.db.get_single_value" no need to get the whole doc

PR #2576, opened 2025-09-11
fix: tab url hash & make a call button

pratikb64:
This PR addresses two bugs:

- Active Tab State Persistence: The active tab was not preserved when the page was reloaded. This has been resolved by syncing the active tab state with the URL hash, ensuring that users return to the same tab after a refresh or navigation.
- "Make a Call" Button Error: Clicking the "Make a call" button was throwing an unhandled error. The underlying issue has been identified and fixed


desk/src/pages/ticket/TicketAgent.vue
```diff
@@ -74,7 +84,22 @@ provide(
   ActivitiesSymbol,
   computed(() => ticketComposable.value.activities)
 );
-
+provide("makeCall", () => {
+  if (
+    !ticketComposable.value.contact.data?.mobile_no &&
+    !ticketComposable.value.contact.data?.phone
```
RitvikSardana, line 91: it should be either,
even if one is set

RitvikSardana (commented):
Pending

PR #2605, opened 2025-09-21
Prevent auto-reopening of closed/resolved tickets on new customer replies

MostafaKadry:
When a customer sends a new reply, the system currently reopens the ticket unconditionally. This leads to Closed or Resolved tickets being reopened automatically, which causes noise and incorrect ticket lifecycle tracking.

1. This PR updates on_communication_update to:
2. Skip auto-status change if the ticket is in Closed or Resolved state.
3. Preserve the integrity of final ticket states unless explicitly reopened by an agent.

### Impact

Prevents unnecessary reopening of tickets that were already resolved. when sending messages via system notifications 

Testing
Add notification and link it with HD Ticket, and set alert on value chage and select field of `status` and just but condition of doc.status in ["Resolved", "Closed"].
then -> Mark Ticket as Resolved/Closed ticket → status should become Resolved/Closed. and remain as same after system send mail notification. (note the issues before is when making this scenario and send mail notification, the ticket re-back to become replayed after it marked as Resolved or Closed because it consider system notification as a new replay).


MostafaKadry:
OK, I overlooked that by mistake!

PR #2623, opened 2025-09-26
fix: ignore automated system notifications in ticket communication from automatically reopens the ticket.

MostafaKadry:
When configuring a Notification For HD Ticket with condition:

doc.status in ["Resolved", "Closed"]

the notification is sent correctly, but it also appears in the Activity as a new Reply, which automatically reopens the ticket and changes its status to Replied.

Expected: Ticket should remain Resolved/Closed after sending the notification.
Actual: Ticket reopens due to the notification being logged as a reply.


MostafaKadry:
@RitvikSardana 
 this pull request that I mentioned in this bug 
https://github.com/frappe/helpdesk/issues/2579

PR #2634, opened 2025-09-30
fix: sqlite support

akhilnarang:
Basically involves setting the correct types here.

    Some date fields had datetime values set, which MariaDB "accepts",
    but won't work with the way we have setup some types for SQLite.

### DIFF
--- a/helpdesk/helpdesk/doctype/hd_service_level_agreement/hd_service_level_agreement.py
+++ b/helpdesk/helpdesk/doctype/hd_service_level_agreement/hd_service_level_agreement.py
@@ -205,7 +205,7 @@ class HDServiceLevelAgreement(Document):
             doc.resolution_date = None
             doc.resolution_time = None
             return
-        doc.resolution_date = now_datetime()
+        doc.resolution_date = frappe.utils.nowdate()
         start_at = doc.service_level_agreement_creation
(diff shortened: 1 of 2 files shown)


helpdesk/helpdesk/doctype/hd_service_level_agreement/hd_service_level_agreement.py
```diff
@@ -355,11 +355,21 @@ def calc_time(
                 current_datetime, current_date
             )
             start_time = max(
-                current_workday_doc.start_time.total_seconds(), current_time_in_seconds
+                timedelta(
+                    hours=current_workday_doc.start_time.hour,
+                    minutes=current_workday_doc.start_time.minute,
+                    seconds=current_workday_doc.start_time.second,
+                ).total_seconds(),
```
RitvikSardana, line 362: what's the difference here?

akhilnarang: `start_time` is a `datetime.time` field, it won't have `total_seconds()`, only `datetime.timedelta` has that.

RitvikSardana:
A lot of changes are made im SLA since July, and relevant test cases are also added.

PR #2666, opened 2025-10-11
fix(ux): init config store from boot when possible

dj12djdjs:
### Summary:
Initializes HD Config into boot context.

The reason I added this is because when the helpdesk page loads, it shows the default helpdesk logo for the first couple of seconds until `helpdesk.api.config.get_config` finishes loading, then it will switch to the branding set in settings.
This change pre loads those settings so there is no switching of the logo after the page has rendered.


niraj2477:
We face the same issue.

RitvikSardana:
can you resolve merge conflicts, so that we can merge this?
@dj12djdjs

Issue #2665, opened 2025-10-10
`helpdesk.search.remove_doc` does not exist


dj12djdjs:
Caused by PR: https://github.com/frappe/helpdesk/pull/2594 @netchampfaris 
https://github.com/frappe/helpdesk/blob/a0df248756e7a9434065590ce09511f0d44dc637/helpdesk/hooks.py#L68

PR #2676, opened 2025-10-17
feat: add Shortcuts to Agent Portal

RitvikSardana:
image

This PR adds a composable written in TS to add shortcuts to the UI.
API:
`useShortcut({ key: "h", meta: true }, ()=>{})`
or
`useShortcut("r", ()=>{})`

This composable handles keys for MacOS and Non MacOS devices.

Shortcuts are disabled when a user is typing in some input element

**Fixes:**
1. Persist router navigation after logging in
2. remove_doc does not exist, closes: #2665 
3. While drafting a response, if the user clicks outside the EmailBox / CommentBox, the box gets hidden

Remove old code for adding shortcuts which never really worked.

### DIFF
--- a/helpdesk/hooks.py
+++ b/helpdesk/hooks.py
@@ -65,8 +65,7 @@ doc_events = {
     },
     "HD Ticket": {
         "on_trash": [
-            "helpdesk.search.remove_doc",
-            "helpdesk.search_sqlite.delete_doc_index",
+            "helpdesk.search_sqlite.delete_doc",
         ],
     },
 }


desk/src/components/modals/ShortcutsModal.vue
```diff
@@ -0,0 +1,128 @@
...
+  description: string;
+}
+
+interface ShortcutGroup {
+  title: string;
+  shortcuts: Shortcut[];
+  hideBorder?: boolean;
+}
+
+const shortcutGroups = computed<ShortcutGroup[]>(() => [
```
surajshetty3416, line 76: One thing that would have made this really cool is if you somehow pulled active shortcuts for the current view.

---

Ideally, your `useShortcut` composable should take "Shortcut Description" and "Group Title" and should have an internal system to list active shortcuts. You can maintain a computed value called `activeShortcuts`, which updates on shortcut attach and destroy.

Modals like this will just use a reactive variable such as `activeShortcuts` to display it.

Once this is done, we'll have a single source of truth and won’t have to maintain a separate map for shortcuts. This can then be part of frappe-ui, as it will be generic enough.

Issue #2714, opened 2025-10-30
Refactor  `apiOptions` on UniInput Component


KerollesFathy:
### **Description:**

When using the **HD Ticket Template**, there’s an option called `url_method` to fetch options for a **Link** field.

The feature works correctly when fetching data, but when the data is returned to the UI, it’s transformed in the `UniInput` component using the following code:

```javascript
const apiOptions = createResource({
  url: props.field.url_method,
  auto: !!props.field.url_method,
  transform: (data) =>
    data?.map((o) => ({
      label: o,
      value: o,
    })) || [],
});
```

The issue is that this transformation assumes each option is a simple string. It doesn’t handle objects that might have distinct `label` and `value` properties.

For example, if the fetched option is a document like `HD-CAT-0001`, it displays that exact string as both the label and value in the dropdown.
This makes it difficult to identify options when you want to show a readable label (e.g., category name) but still use the document name as the value.

---

### **Expected Behavior:**

The `url_method` should support returning objects with `label` and `value` keys, so that the UI can display a readable label while keeping the correct internal value.

---

### **Actual Behavior:**

All returned data is mapped to `{ label: o, value: o }`, ignoring cases where the API returns objects with their own `label` and `value` fields.

---

### **Suggested Refactor:**

Update the `transform` logic to handle both cases, strings **and** objects:

```javascript
transform: (data) =>
  data?.map((o) =>
    typeof o === "object" && o.label && o.value
      ? o
      : { label: o, value: o }
  ) || [],
```

PR #2715, opened 2025-10-30
refactor: update apiOptions transform to map object labels and values

KerollesFathy:
Closes: #2714 

---

After this PR is merged, you can return data in the following format from server side using the `url_method` field in the **Fields** child table of the **HD Ticket Template**:

```python
[
    {"label": "Support Category", "value": "HD-CAT-0001"},
    {"label": "Billing Category", "value": "HD-CAT-0002"}
]
```

The `transform` function in `apiOptions` within the `UniInput` component will now correctly map these objects into dropdown options.

### DIFF
--- a/desk/src/components/TicketField.vue
+++ b/desk/src/components/TicketField.vue
@@ -75,10 +75,14 @@ const apiOptions = createResource({
   transform: (data) => {
     if (!data?.length) return [];
     return (
-      data?.map((o) => ({
-        label: o,
-        value: o,
-      })) || []
+      data?.map((o) =>
+        typeof o === "object" && o.label && o.value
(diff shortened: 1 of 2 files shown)


KerollesFathy:
> Does this support previous implementation as well?
> 
> Please share SS for both scenarios, when API response is
> 
> [{"label": "Support Category", "value": "HD-CAT-0001"}]
> 
> and when the response is
> 
> ["Billing", "Product"]

@RitvikSardana 
Yes, it covers both scenarios. Here’s a video.

https://github.com/user-attachments/assets/6d486b2f-1980-4eb7-b523-01f77e588d6d

PR #2770, opened 2025-11-25
Refactor total_count calculation in doc.py

laxmantandon:
Frappe v16 no longer allows expressions like "count(*) as count" inside frappe.get_list()


RitvikSardana:
will be closed by #2789

PR #2781, opened 2025-11-27
feat: rtl support

RitvikSardana:
Add RTL Support in Helpdesk

telegram-cloud-document-5-6143335016186256458


RitvikSardana:
@aerodeval will work on it, hence closing

PR #2818, opened 2025-12-10
chore: show app version on sidebar

KerollesFathy:
This  helps users quickly see their current Helpdesk app version. When they click the version number, they are taken directly to the release notes on GitHub. This is especially useful when reporting bugs, as users can easily confirm which version they are running.

https://github.com/user-attachments/assets/0fe1d4b7-1b3a-46e5-8864-0ccb1abe1bb0

### DIFF
--- a/desk/src/components/layouts/Sidebar.vue
+++ b/desk/src/components/layouts/Sidebar.vue
@@ -159,6 +159,13 @@
       :currentStep="currentStep"
     />
     <CP v-model="showCommandPalette" />
+    <span
+      v-if="appVersion"
+      class="text-sm text-gray-600 cursor-pointer"
+      @click="goToRelease"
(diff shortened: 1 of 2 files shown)


RitvikSardana:
Hey @KerollesFathy 

We wont be showing the version like this in the sidebar.

In future we will add a "About" tab inside user drop down and user can customize the table just like CRM. / Learning

KerollesFathy:
> Hey @KerollesFathy 
> 
> 
> 
> We wont be showing the version like this in the sidebar.
> 
> 
> 
> In future we will add a "About" tab inside user drop down and user can customize the table just like CRM. / Learning

Thanks @RitvikSardana for clarifying.Okey, I will close the PR.

PR #2821, opened 2025-12-11
ci: setup semantic-release gh workflow

KerollesFathy:


KerollesFathy:
> It creates a new tag? does it update init file also?

Yes, and update release note

PR #2827, opened 2025-12-15
fix: use get_list to get count

RitvikSardana:
**Issue**
get_count does not apply permissions, hence resulting in incorrect count.

**Solution**
Use get_list to get the count


rehanrehman389:
closes https://github.com/frappe/helpdesk/issues/2825

Issue #2838, opened 2025-12-22
Reports in the workspace aren’t routing correctly


KerollesFathy:
https://github.com/user-attachments/assets/16bfaddf-5d6a-48cf-86a4-634f18c310ca

PR #2839, opened 2025-12-22
fix(workspace): helpdesk reports routing

KerollesFathy:
**Closes:** #2838  

---
https://github.com/user-attachments/assets/33c005d9-c9fa-4ea9-9314-863676a80ebc


RitvikSardana:
The whole JSON along with indentation has changed, could you fix this to only show the changed made?

Issue #2851, opened 2025-12-31
UX: Mentions should be `bold`


ankush:
<img width="550" height="169" alt="Image" src="https://github.com/user-attachments/assets/422ace08-9d03-49ee-9db0-c5fc60447d09" />

Right now I can't tell if mention actually worked or not. They should be bold + linked to user (?)

RitvikSardana:
Oh, this seems to be some kind of regression, will fix

vkWeb:
@RitvikSardana @ankush I've fixed and raised a PR (https://github.com/frappe/helpdesk/pull/2867). Hoping to see it get merged soon 🙌

PR #2867, opened 2026-01-09
fix: rendering @mention in comments

vkWeb:
Fixes https://github.com/frappe/helpdesk/issues/2851

The @mention in comments were broken (the below image is from #2851):

531230868-422ace08-9d03-49ee-9db0-c5fc60447d09

This PR fixes that by passing the `mentions` prop to the TextEditor. It results in correct rendering of @ mention span element.

image


RitvikSardana:
Thank you for the PR @vkWeb 

The solution is correct.

PR #2890, opened 2026-01-15
fix: enable keyboard navigation (Page Up/Down, Home, End) on tickets list

### LINKED ISSUE #2112: Keyboard navigation doesn't work
### Information about bug

Page Up / Down, Home and End doesn't work.

### Version

develop

### Installation method

None

_No response_

suraniharsh:
- Add useListScroll composable to handle Page Up/Down, Home, End keys
- Apply to tickets list page to enable scrolling with keyboard
- Fixes #2112

https://github.com/user-attachments/assets/29dabf79-6cd3-416a-90a8-694b7e1f42b5

### DIFF
--- a/desk/src/composables/shortcuts.ts
+++ b/desk/src/composables/shortcuts.ts
@@ -33,7 +33,11 @@ export const useShortcut = (
 
     if (validateKeyCombination(e, shortcutBinding)) {
       try {
-        e.preventDefault();
+        // Don't prevent default for page navigation keys
+        const navigationKeys = ['PageUp', 'PageDown', 'Home', 'End'];
+        if (!navigationKeys.includes(e.key)) {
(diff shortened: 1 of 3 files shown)


RitvikSardana:
@suraniharsh 
I think, It should be fixed in the ListView component of frappe-ui

suraniharsh:
> @suraniharsh I think, It should be fixed in the ListView component of frappe-ui

Good point  makes sense to handle this at the `ListView` level instead of patching it on the tickets page.

I’ll refactor the implementation and move the keyboard navigation logic into `frappe-ui`’s `ListView` component so it works consistently across all lists.

RitvikSardana:
let us close this PR then @suraniharsh ?

suraniharsh:
Yes,
lets close this PR

PR #2904, opened 2026-01-17
ci: fix generate POT workflow

earona:
This pull request updates the GitHub Actions workflow for generating POT files by adding a step to set up Node.js before running the update script.

### DIFF
--- a/.github/workflows/generate-pot-file.yml
+++ b/.github/workflows/generate-pot-file.yml
@@ -27,6 +27,11 @@ jobs:
         with:
           python-version: "3.14"
 
+      - name: Setup Node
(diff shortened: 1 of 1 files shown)


RitvikSardana:
What would be fixed by this?

mahsem:
> Hello @RitvikSardana, This PR fixes the bug in the GitHub action that generates translation strings for Crowdin.
> 
> This bug was introduced on December 24, 2025, when the Node.js requirements for the framework were updated.
> 
> Ref: https://github.com/frappe/helpdesk/actions/workflows/generate-pot-file.yml

@RitvikSardana can this be merged

PR #2960, opened 2026-02-02
fix: remove unwanted files

RitvikSardana:


helpdesk/api/assignment_rule.py
```diff
@@ -3,30 +3,27 @@
...
+            frappe.PermissionError,
         )
-        users_exists = bool(
-            frappe.db.exists("Assignment Rule User", {"parent": docname.name})
+
+    assignment_rules = frappe.get_all(
+        "Assignment Rule",
+        fields=["name", "description", "disabled", "priority"],
+        order_by="modified desc",
+    )
```
sagarvora, line 16: use get_list. does permission checks and queries. check can continue to be explicit also - if you want the custom error message, but get_list should still be done for the permission query.

helpdesk/api/assignment_rule.py
```diff
@@ -3,30 +3,27 @@
...
+
+    parents_with_users = set(
```
sagarvora, line 18: this is fine as get_all because you're filtering on permitted records from parent table

helpdesk/api/saved_replies.py
```diff
@@ -6,11 +6,10 @@
 
 @frappe.whitelist()
 @agent_only
-def get_rendered_saved_reply(ticket_id, saved_reply_id=None, saved_reply=None):
-    if not (saved_reply_id or saved_reply):
-        frappe.throw(_("Please provide either saved_reply_id or saved_reply data"))
-    if not saved_reply:
-        saved_reply = frappe.get_doc("HD Saved Reply", saved_reply_id).message
+def get_rendered_saved_reply(ticket_id, saved_reply_id=None):
```
sagarvora, line 9: add type hints. pydantic enforces them

helpdesk/helpdesk/doctype/hd_ticket/api.py
```diff
@@ -41,7 +41,7 @@ def new(doc, attachments=[]):
 
 @frappe.whitelist()
 def get_one(name, is_customer_portal=False):
```
sagarvora, line 43: type hints

helpdesk/helpdesk/doctype/hd_ticket/api.py
```diff
@@ -759,6 +767,12 @@ def get_similar_tickets(ticket: str):
         },
         as_dict=1,
     )
+    tickets = [
+        t for t in tickets if frappe.has_permission("HD Ticket", "read", doc=t["name"])
+    ]
```
sagarvora, line 772: this will do get_doc internally for each doc - not good

sagarvora: use qb.get_query with ignore_permissions=False and then add your filters

PR #2994, opened 2026-02-09
fix: hide fields in customer portal based on field dependency

### LINKED ISSUE #2993: Field Dependency should hide the field in customer portal
### Current Behaviour:
Fields configured to be hidden via Field Dependency are still visible in the customer portal.

### Expected Behaviour:
Fields with dependency criteria that evaluate to false should be hidden in the customer portal, just like in the agent/desk view.

### Steps to Reproduce:

- Set up a Field Dependency to hide a field based on another field’s value.
- Log in to the customer portal.
- The field is still visible, even when it should be hidden.

Image

> Customer portal view: 

Image

### Version

Frappe version- 15
Frappe Helpdesk version - main

### Installation method

manual install

```shell
NO stacktrace
```

trufurs:
Closes #2993

This PR adds eval method for `display depends on` on the customer portal using the same function used in Agent portal

### DIFF
--- a/desk/src/pages/ticket/TicketCustomerTemplateFields.vue
+++ b/desk/src/pages/ticket/TicketCustomerTemplateFields.vue
@@ -51,6 +51,7 @@ import { dayjs } from "@/dayjs";
 import { Field } from "@/types";
 import { computed, inject } from "vue";
 import { ITicket } from "./symbols";
+import { parseField } from "@/composables/formCustomisation";
 
 const ticket = inject(ITicket);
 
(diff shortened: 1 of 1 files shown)


RitvikSardana:
@trufurs 

This is not complete, if we are going to implement Field Dependency on the ticket view, we should implement it in Agent Portal as well.


It should filter the select fields / link fields similar to how it is done in the "New Ticket View"

trufurs:
@RitvikSardana 
It is there on agent portal, see the sub category is hidden automatically
image

trufurs:
@RitvikSardana 
i see this field dependency issue  of not using `display on`  on agent portal -mobile view as well but it currently works on agent portal desktop view
so should i add 2 fields 
Apply Field Dependency in Customer View
Apply Field Dependency in Agent View

or  
Apply Field Dependency in Ticket View -> only
and i also change the current desktop view to evaluate only if this is enabled ?

can you please guide me with a simple requirements and what should be the end result ?
also i started working on filter table dependency for parent child in agent portal ( i will make separate PR for it , or if you want in one i can do that)

PR #3006, opened 2026-02-12
fix: comment box mentions bug on edit

aerodeval:
Previously when comments were edited, mentions feature didn't work correctly

https://github.com/user-attachments/assets/4dba0bd3-0d28-4a0c-80d9-d9e45427e7ed

after fix:
https://github.com/user-attachments/assets/d51a936a-0b97-49b1-a8f8-f75dbd1e0e14

additionally also, the notifications sent in notify_mentions is refactored as it would send multiple notifications i.e to both users whenever the comment is edited which does not make sense.

on further testing we found a bug in discard button which had weird bug affecting mentions to not parse in tiptap.

https://github.com/user-attachments/assets/095a2c8a-8549-46b1-ba8b-4bf2b706e7fb

this issue is fixed.

closes: https://github.com/frappe/helpdesk/issues/2650

### DIFF
--- a/desk/src/components/CommentBox.vue
+++ b/desk/src/components/CommentBox.vue
@@ -64,10 +64,24 @@
         :bubble-menu="textEditorMenuButtons"
         :mentions="userMentions"
         @change="(event:string) => {_content = event}"
+        @keydown.ctrl.enter.capture.stop="handleSaveComment"
+        @keydown.meta.enter.capture.stop="handleSaveComment"
       >
         <template #bottom v-if="editable">
           <div class="flex flex-row-reverse gap-2">
-            <Button label="Save" @click="handleSaveComment" variant="solid" />
+            <div>
+              <Button
+                :label="
+                  isMobileView
+                    ? 'Save'
+                    : isMac
+                    ? 'Save (⌘ + ⏎)'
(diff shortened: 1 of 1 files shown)


desk/src/components/CommentBox.vue
```diff
@@ -258,6 +261,10 @@ function handleDiscard() {
   editable.value = false;
 }
 
+const mentionsForEditor = computed(() => {
+  return dropdown.value;
+});
```
RitvikSardana: why not just use the dropdown value directly, and rename the dropdown variable?

aerodeval: It wasn't getting rendered properly while I was testing so I added it for safety. will make the changes

desk/src/components/CommentBox.vue
```diff
@@ -205,6 +207,7 @@ const isTicketMergedComment = computed(() => {
   const regex = /has been merged with ticket #\d+/;
   return regex.test(content);
 });
+const { dropdown } = storeToRefs(useAgentStore());
```
RitvikSardana: what is the need of storeToRefs?

you can just use this `drowdown` variable in the mentions prop.

Why create a separate variable?

PR #3009, opened 2026-02-12
feat: time localisation 

aerodeval:
This pr fixes issues caused in time calculation of ticket created on, resolution, etc. Issue was faced whenever user had different timezone than timezone of the system settings.
image

after fix:
image

also tool tip displays now time relative to the user timezone as earlier it was confusing and not reflecting according to that

user in IST timezone 
Screenshot 2026-02-20 at 5 15 54 PM

user viewing same ticket in usa EST timezone
image

also allows agent to switch timezone from profile modal

image

the fix now displays all time according to local timezone

closes https://github.com/frappe/helpdesk/issues/2487  and https://github.com/frappe/helpdesk/issues/1688


desk/src/utils.ts
```diff
@@ -52,11 +54,143 @@ export function validateEmailWithZod(email: string) {
...
+  return Intl.DateTimeFormat().resolvedOptions().timeZone
+}
+
+export function prettyDate(date, mini = false) {
+  if (!date) return ''
+
+  let systemTimezone = getConfig('systemTimezone')
+  let localTimezone = getConfig('localTimezone') || getBrowserTimezone()
+  if (typeof date == 'string') {
+    date = dayjsLocal(date)
```
RitvikSardana, line 73: use dayjsLocal directly, no need to get the config and set the TZ again and again

desk/src/components/ListViewBuilder.vue
```diff
@@ -136,12 +136,13 @@ import { capture } from "@/telemetry";
...
 import { formatTimeShort, getIcon } from "@/utils";
 import { useStorage } from "@vueuse/core";
-
+import dayjs from "dayjs";
 import { useTicketStatusStore } from "@/stores/ticketStatus";
 import { __ } from "@/translation";
 import {
   call,
   createResource,
+  dayjsLocal,
```
RitvikSardana, line 144: remove unused imports

RitvikSardana (changes requested):

desk/src/components/ticket/TicketCustomerSidebar.vue
```diff
@@ -87,13 +87,19 @@
       <div
         class="flex items-center text-base leading-5"
         v-for="field in ticketAdditionalInfo"
+        :key="field.label"
```
RitvikSardana: labels can be same for fields, fieldname is unique

RitvikSardana (changes requested):

PR #3050, opened 2026-02-23
fix: email send button bug and null value appearing in saved replies

aerodeval:
image

in Email Area, occasionally cache  would store an object null thus allowing send button to be enabled even when empty also when using saved replies garbage value was added

image

also a bug where replied content doesn't come under collapsed
image

fix:
image

one more bug where on reply the send button is still disabled or has garbage value at times

https://github.com/user-attachments/assets/56cc2aa0-33fd-4f4b-820e-bc0949fd8680

this fixes all of these issues


RitvikSardana:
Remove newly added computed property

PR #3120, opened 2026-03-13
Fixed the remove button issue that worked like a hyperlink

shafineit:


RitvikSardana:
If you raise one more PR like this, you will be banned from the repo @shafineit

PR #3134, opened 2026-03-18
fix: prevent saved replies to run queries in modal selector view

aerodeval:
sanitize input text in saved replies


aerodeval:
frappe.render_template is being reworked on by internal team so the issue will be fixed soon and the above changes will not be needed

PR #3137, opened 2026-03-20
fix: set Link field page limit to 999

regdocs:


desk/src/components/UniInput.vue
```diff
@@ -68,6 +68,7 @@ const component = computed(() => {
     return h(Link, {
       doctype: props.field.options,
       filters: props.field.filters,
+      pageLength: getCurrentInstance()?.parent?.type?.__name === 'TicketNew' ? 999 : 10,
```
RitvikSardana: we are using this component only in `TicketNew`  page only

Image

so we can directly use 999

PR #3151, opened 2026-03-29
fix: add mandatory agent check for agent actions

aerodeval:
Add check to assign agent function when  an agent is being assigned to a ticket

### DIFF
--- a/helpdesk/helpdesk/doctype/hd_ticket/hd_ticket.py
+++ b/helpdesk/helpdesk/doctype/hd_ticket/hd_ticket.py
@@ -257,7 +257,7 @@ class HDTicket(Document):
         self.ticket_type = ticket_type
 
     def set_raised_by(self):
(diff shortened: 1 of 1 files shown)


helpdesk/helpdesk/doctype/hd_ticket/hd_ticket.py
```diff
@@ -256,7 +257,7 @@ def set_ticket_type(self):
         self.ticket_type = ticket_type
 
     def set_raised_by(self):
-        self.raised_by = self.raised_by or frappe.session.user
+        self.raised_by = self.raised_by if is_agent() else frappe.session.user
```
RitvikSardana, line 260: why?

aerodeval: allowing only agents to change and if not agent defaulting to session user always

PR #3208, opened 2026-04-08
feat: added column type change migration

### LINKED ISSUE #3207: Migrate naming series of HD Ticket from auto-increment to custom naming series
Currently, HD Ticket uses auto-increment integers as document names (e.g., 1, 2, 3). This makes ticket IDs non-descriptive and harder to reference in external communications.

This issue tracks migrating the naming to a human-readable custom naming series such as HD-TKT-.YYYY.-.##### (e.g., HD-TKT-2026-00001).

ToDo
- [ ] Update in HD Ticket DocType from autoincrement to a naming series
- [ ] Ensure all references to ticket.name handle string names correctly (no parseInt assumptions)

dankgarlic1:
Moving the ticket naming system from simple auto-incrementing integers to a structured, year-based format: HD-TKT-YYYY-#####.

The Architectural Approach
I split this into a two-phase migration to ensure the database stays stable:

Phase 1: Schema Alignment
The first patch converts the name column from BigInt to VarChar. Since we are moving from numbers to strings, the "pipe" had to be widened first before we could push any new data through.

Phase 2: Data Transformation & Seeding
The second patch handles the heavy lifting:

Backfilling: It iterates through legacy tickets and renames them to the new format based on their original creation year.

Series Recovery: It calculates the highest existing ID for each year to seed the tabSeries table. This ensures the next ticket created in the UI follows the sequence perfectly without crashing.

Performance Note
For the series seeding, the logic looks a bit less readable and ugly, but I thought if I did it in Python the OOM could happen. Not sure what kind of volumes the Helpdesk users have, so I went with SQL to keep the RAM safe by letting the DB handle the grouping and max-ID calculations.

fixes #3207

### DIFF
--- a/helpdesk/patches/backfill_legacy_hd_tickets.py
+++ b/helpdesk/patches/backfill_legacy_hd_tickets.py
@@ -3,6 +3,11 @@ from frappe.utils import getdate
 
 
 def execute():
+    migrate_ticket_names()
+    seed_series
+
+
+def migrate_ticket_names():
     legacy_tickets = frappe.db.sql(
         """
         SELECT name, creation 
@@ -33,6 +38,10 @@ def execute():
 
         frappe.db.commit()
 
+
+def seed_series():
+    # Extract year and sequence from names like:
+    # HD-TKT-2026-00001
     yearly_max_data = frappe.db.sql(
         """
         SELECT 
--- a/helpdesk/patches/backfill_legacy_hd_tickets.py
+++ b/helpdesk/patches/backfill_legacy_hd_tickets.py
@@ -1,69 +1,70 @@
 import frappe
 from frappe.utils import getdate
 
+
 def execute():
-    # 1. FETCH LEGACY TICKETS
-    # We use NOT LIKE to skip any tickets already renamed in previous tests
-    legacy_tickets = frappe.db.sql("""
+    legacy_tickets = frappe.db.sql(
+        """
         SELECT name, creation 
         FROM `tabHD Ticket` 
         WHERE name NOT LIKE 'HD-TKT-%'
         ORDER BY CAST(name AS UNSIGNED) ASC
-    """, as_dict=True)
+    """,
+        as_dict=True,
+    )
 
     if legacy_tickets:
-        # 2. RENAME LOOP
         for index, ticket in enumerate(legacy_tickets):
             old_name = ticket.name
             try:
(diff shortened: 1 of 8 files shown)


helpdesk/patches/backfill_legacy_hd_tickets.py
```diff
@@ -0,0 +1,70 @@
...
+        """
+        SELECT 
+            SUBSTRING(name, 8, 4) as year, 
+            MAX(CAST(RIGHT(name, 5) AS UNSIGNED)) as max_id
+        FROM `tabHD Ticket`
+        WHERE name LIKE 'HD-TKT-____-_____'
+        GROUP BY year
+    """,
+        as_dict=True,
+    )
```
dankgarlic1, line 55: The series seeding logic looks a bit less readable and ugly, but I thought if I did it in Python the OOM could happen. Not sure what kind of volumes the Helpdesk users have, so I went with SQL to keep the RAM safe

dankgarlic1:
@RitvikSardana please take a look at this

RitvikSardana:
So had some discussions, and we will go ahead with the default value of naming series as `####`.

For existing tickets, no need to change to alter the column type and update legacy tickets, we will take the highest ticket count, and set it as the counter in the DB via patch.
So in DB, naming series maintain a count, and we can update that count via patch to the biggest ticket ID.
If the biggest ticket ID is 59000.
The counter for the naming series will be adjusted to 59,000 so that when the new ticket is created, it will start from 59,001.

For new site the name will then automatically start form 0001.
@dankgarlic1 

Once this is done, let me know.

dankgarlic1:
@RitvikSardana closing this PR, I made a new PR for this change. You can review it here #3276 

> So had some discussions, and we will go ahead with the default value of naming series as `####`.
> 
> For existing tickets, no need to change to alter the column type and update legacy tickets, we will take the highest ticket count, and set it as the counter in the DB via patch. So in DB, naming series maintain a count, and we can update that count via patch to the biggest ticket ID. If the biggest ticket ID is 59000. The counter for the naming series will be adjusted to 59,000 so that when the new ticket is created, it will start from 59,001.
> 
> For new site the name will then automatically start form 0001. @dankgarlic1
> 
> Once this is done, let me know.

@RitvikSardana  closing this PR, I made a new PR for this change. You can review it here #3276

PR #3211, opened 2026-04-10
fix: add clear formatting utility in text editor

aerodeval:
Implemented a formatting utility in text editors

https://github.com/user-attachments/assets/810faf00-c7a3-4389-82e8-74cdde89ff8a

inspired by the formatting feature in drive

**why**
Support agents constantly paste content from external sources (email, docs, web), which brings in inline styles, classes, and inconsistent markup. As a result, they can use this shortcut to clean up faster.

Other fixes:

- better handling for rtl based inputs in text editor
i.e 
image

- better handling of pre tag

before :
image

after:
image

### DIFF
--- a/desk/src/components/CommentTextEditor.vue
+++ b/desk/src/components/CommentTextEditor.vue
@@ -14,7 +14,7 @@
     :editable="editable"
     :mentions="dropdown"
     @change="editable ? (newComment = $event) : null"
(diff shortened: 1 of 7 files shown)


RitvikSardana:
Shouldn't we also do it for customer text editor @aerodeval ?

PR #3214, opened 2026-04-13
Allow only a group of agents to have a assign permission

Shindhu-Ramaswamy:
Currently, any user can assign or remove assignees from a ticket.
There is no restriction based on roles.

This can cause issues because:

Unauthorized users can change ticket assignments
No proper control over who can manage tickets

For that i came up with the solution:

I added a role-based check to control ticket assignment.

Backend
Created a method can_assign_ticket() to check user roles
Allowed roles:
Ticket Assigner
Agent Manager
System Manager
Added validation in remove_assignments:
If user does not have permission → throw error
Frontend
Updated AssignTo.vue
If user has permission:
Show assign UI
If not:
Show assignee list as read-only

#issue https://github.com/frappe/helpdesk/issues/2488

### DIFF
--- a/desk/src/components/ticket-agent/AssignTo.vue
+++ b/desk/src/components/ticket-agent/AssignTo.vue
@@ -1,4 +1,5 @@
 <template>
+  <template v-if="canAssign">
   <Popover class="flex w-full" placement="bottom-end">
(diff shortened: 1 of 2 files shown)


RitvikSardana:
Not planned as of now, hence closing

PR #3224, opened 2026-04-21
fix(ticketAgent): sidepanel and reply icons

RitvikSardana:
1. Change the TextArea to InputField in the side panel of the ticket view
2. Change width of reply, reply all icons
**Before:**
image

**After:**
image
3. Refactor sidebar design
image
4. Remove Activity Header "New" dropdown.
5. Use Ghost Badge for showing SLA in TicketView


desk/src/components/EmailArea.vue
```diff
@@ -50,13 +50,21 @@
...
+            :icon="ReplyIcon"
+            @click="reply"
+          />
+          <Button
+            :tooltip="__('Reply All')"
+            variant="ghost"
+            :icon="ReplyAllIcon"
+            class="!text-ink-gray-7"
+            @click="replyAll"
+          />
```
netchampfaris, line 67:
```html
<Button
  :tooltip="__('Reply All')"
  variant="ghost"
  @click="replyAll"
>
    <ReplyAllIcon class="text-ink-gray-7" />
</Button>
```

Correct way to do it without using ! in classname

RitvikSardana: The code was like this only due to which there were width issues.
@netchampfaris 

image
First 3 are with new code, next 3 are with the way you mentioned.

PR #3239, opened 2026-04-23
fix: hide onboarding banner and popup after onboarding is completed

aerodeval:
the Help Center banner is always visible in the menu even after user is familiar with the product which is sometimes annoying

image

this pr aims to fix that


RitvikSardana:
Closing this PR; reopen it whenever you continue working on it

PR #3241, opened 2026-04-23
fix: mobile ui issues

aerodeval:
Fix mobile view issues where text is misaligned and Agent responses are missing


michellealva:
fixes: https://github.com/frappe/helpdesk/issues/3240

PR #3266, opened 2026-04-30
fix: allow sending with bcc/cc

### LINKED ISSUE #3254: Allow sending mails with no "To:" entries, but filled "CC:" or "BCC:" entries
### Information about bug

When sending a reply / new ticket, after emptying the "To:" field, but adding entries to CC or BCC, you can't send the email, as the "To:" field isnt filled.

In our case, we use this approach to send mails to multiple customers at the same time (while using the BCC to obviously not leak email addresses to everyone), so that we can catch any replies via the helpdesk.

### Version

```json
{
	"frappe": "16.11.0",
	"telephony": "0.0.1",
	"helpdesk": "1.21.1"
}
```

### Installation method

easy-install

```shell

```

aerodeval:
Currently, we do not allow mails to be sent when `To` field is empty and other fields like CC and BCC are specified. 

This PR fixes the following issue

Allow sending mails with no "To:" entries, but filled "CC:" or "BCC:" entries

image

i.e system should allow sending mails like this

image
in email view

image
in communications (to is empty)

closes: https://github.com/frappe/helpdesk/issues/3254


desk/src/components/EmailArea.vue
```diff
@@ -173,6 +174,32 @@ const status = computed(() => {
   return { label: _status, color: indicator_color };
 });
 
+const normalizeAndFilter = (
+  field: string | string[],
+  filterOut: string[] = []
```
RitvikSardana: better naming would be, valuesToExclude, or excludedEmails

helpdesk/helpdesk/doctype/hd_ticket/test_hd_ticket.py
```diff
@@ -705,6 +706,55 @@ def test_if_banner_not_shown_after_next_working_day(self):
...
 
+    def test_reply_via_agent_with_only_cc(self):
+        """
+        reply_via_agent should succeed when only cc is provided and to is empty/None
+        """
+        ticket = make_ticket()
+        cc_recipient = "cc_only@test.com"
+
+        # Should not raise a validation error
+        try:
```
RitvikSardana: instead of this you can also check if the communication created, and the CC is empty

helpdesk/helpdesk/doctype/hd_ticket/test_hd_ticket.py
```diff
@@ -705,6 +706,55 @@ def test_if_banner_not_shown_after_next_working_day(self):
...
+        ticket = make_ticket()
+        bcc_recipient = "bcc_only@test.com"
+
+        try:
+            ticket.reply_via_agent(message="Test reply", to=None, bcc=bcc_recipient)
+            sent = True
+        except Exception:
+            sent = False
+
+        self.assertTrue(sent)
```
RitvikSardana: same as above

RitvikSardana (dismissed):

PR #3329, opened 2026-05-10
fix: UI bugs 

aerodeval:
UI Bugs faced due to some Frappe UI changes and some improvements in UX.

settings with a toggle button are saved directly when enabled
image

dropdown & combobox fixes to use new slots

Widths for Link fields were messed up causing ui bugs
image

timespan now uses autocomplete instead of select component which caused a long list to show up

image

dropdown now appears properly in breadcrumbs while pinning or duplicating a view
image

centered loading states in the settings panel and fixed a few ui bugs like

image

outline getting cut off


RitvikSardana:
Use Combobox component instead of Autocomplete @aerodeval

PR #3347, opened 2026-05-12
refactor: Use read replica for all get APIs

niraj2477:
This pr introduces 2 changes:

- Wrap GET calls with `frappe.read_only`.
- Allow anyone to create ticket via `helpdesk.doctype.hd_ticket.api.new`.
    -  There is setting in `HD Settings` which allows anyone to create the ticket if setting is enabled, but the api itself does not checks the guest permission, so check is added for same.

ref: https://github.com/frappe/helpdesk/issues/3336

### DIFF
--- a/helpdesk/api/agent_home/agent_home.py
+++ b/helpdesk/api/agent_home/agent_home.py
@@ -17,6 +17,7 @@ from helpdesk.utils import agent_only, format_time_difference
 
 
 @frappe.whitelist()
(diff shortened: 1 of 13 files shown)


RitvikSardana:
Follow conventional commits for PR title aswell

niraj2477:
Updated

niraj2477:
Closing this, will be raising new one instead.

PR #3361, opened 2026-05-15
feat: sync ERPNext customer with HD customer

RitvikSardana:
**Summary**
This PR adds bidirectional customer sync between Frappe Helpdesk and ERPNext. 
When the integration is enabled, creating/updating/renaming an HD Customer mirrors changes to an ERPNext Customer record, and vice versa. 

**How it works**
We have a new settings doctype called `ERPNext HD Settings`.
This can only be enabled if both ERPNext & Helpdesk is installed on the same site.
If disabled => The two masters "ERPNext Customer" & "HD Customer" will work separately, there will be no syncing.

If enabled => 
A button is shown on both Customer List View in the frontend of Helpdesk & Customer List view in ERPNext (desk).

> You need to have perms to write in both the doctypes to perform this action.

This is what happens when the button is pressed:

1. A job is enqueued to sync the two masters
So let us say if

ERPNext Customer has:
  - ERP A
  - ERP B
  - Common

HD Customer has:
  - HD A
  - HD B
  - Common

After the job is finished both will have

  - ERP A
  - ERP B
  - Common
  - HD A
  - HD B

> Same name masters are ignored.

> User Perms & Shared docs are also synced when the job is finished.

**What Changed** 
1. New ERPNext HD Settings singleton DocType
3. `erpnext_customer` field on HD Customer
4. `hd_customer` field in ERPNext Customer DocType.
5. Background-job-based bulk sync, a settings toggle in the Helpdesk UI, and a sync button on the Customers list page.
6. Added test cases related to the integration.

`docs`: https://docs.frappe.io/helpdesk/integration/erpnext


michellealva:
Issues:

- [x] Rename /  merge customer doesn't rename HD customer
- [x] User permissions need to be honoured in both the app
- [x] When a customer record is shared in one app, also share it in the other
- [x] Initial sync doesn't copy the customer image
- [x] When there are an equal number of records in both apps, the sync button isn't shown
- [x] Sync button not shown in ERPNext Customer's List View
- [x] If there are linked documents in one app, do not let me delete the customer record in another app 
- [x] Customer image isn't shown in list view.

PR #3367, opened 2026-05-18
fix: correct label for SLA when ticket is on hold

RitvikSardana:
**Issue:**
Incorrect label in Resolution By when ticket is On Hold

`depends-on`: https://github.com/frappe/helpdesk/pull/3297


RitvikSardana:
@mergifyio queue

PR #3372, opened 2026-05-19
feat: agent  availability status

### LINKED ISSUE #2872: Enhance Agent's Profile
1. Add status field in HD Agent doctype for Available / Away
    If the Agent is away the ticket should not be assigned to the Agent.

2. Add Agent's Signature, when the Ticket is mounted, automatically add Agents's signature, in the Reply Section.
This will be overridden if the Agent uses Canned Responses / Saved Replies.

aerodeval:
# Introducing Agent Availability Status

https://github.com/user-attachments/assets/9c10ba54-f0fc-40fa-a1e1-063f346713a7

This feature lets agents set their availability status so the helpdesk knows who can actually pick up work. Agents update their status from their profile menu, and assignment rules respect it when handing out tickets.

## How it works

Every status belongs to one of three categories: **Active**, **Away**, or **Unavailable**. When a ticket is assigned through an assignment rule (Round Robin, Load Balancing, or Weighted Distribution), we filter the rule's user pool by availability.

Assignment priority works as follows:

**Active → Away → Unavailable**

So if two agents are Away and one is Active, the ticket always goes to the Active agent. Away agents are only considered when no Active agent is in the pool, and Unavailable agents are used only as a last resort, so a ticket is never left unassigned just because everyone stepped out. If nobody in the pool has set an Away or Unavailable status, we fall through to the framework's default behaviour, so nothing changes for teams that don't use this.

Manual assignment still works the same way. If you assign a ticket to someone who's Away or Unavailable, you get a toast warning, but the assignment goes through.

## Custom statuses

image

Teams aren't limited to the three defaults. You can create your own statuses and slot them under whichever category fits. Above, **Vacationing** is a custom status under the Unavailable category. Each status has a configurable color (shown as a presence dot next to the agent) and an order that controls how it appears in the menu.

### The three categories

- **Active**: currently working and can be assigned tickets.
- **Away**: on a short break or stepped away, can't pick up a ticket right now.
- **Unavailable**: out entirely and can't access tickets at all.

Every status maps to one of these, so organisations can model their own workflows while assignment logic stays consistent.

## Important

There's always exactly one Active status, and it can't be disabled, deleted, or moved to another category. It's the guaranteed fallback that keeps assignment working. You can add as many Away and Unavailable statuses as you need.

docs: https://docs.frappe.io/helpdesk/other-features/agent-availability-status

closes https://github.com/frappe/helpdesk/issues/2872, https://github.com/frappe/helpdesk/issues/2824


helpdesk/overrides/assignment_rule.py
```diff
@@ -0,0 +1,54 @@
...
+
+        away = get_away_agents()
+        if not away:
+            return super().get_user(doc)
+
+        original_users = self.users
+        original_weighted = self.weighted_users
+
+        available_users = [u for u in original_users if u.user not in away]
+        available_weighted = [u for u in original_weighted if u.user not in away]
```
RitvikSardana: could be done according to the assignment rule type which is set

aerodeval: https://github.com/frappe/frappe/commit/115d3cead06fb2a5ed75c2713f4d31c9ede3531c#diff-7ecd563e403db1a550780aee4256d5bba511abb8eccf26a78dc376f2533a5f39R40 

weighted distribution is a separate child table hence a different variable used for that here, but yes no point in making this variable if the assignment rule is not that

helpdesk/helpdesk/doctype/hd_agent_status/test_hd_agent_status.py
```diff
@@ -0,0 +1,120 @@
...
+    def _make_status(self, agent_status, category="Away", enable=1, order=None):
+        return frappe.get_doc(
+            {
+                "doctype": "HD Agent Status",
+                "agent_status": agent_status,
+                "category": category,
+                "enable": enable,
+                "order": order,
+            }
+        ).insert()
```
RitvikSardana: should be inside test_utils

helpdesk/helpdesk/doctype/hd_agent_status/test_hd_agent_status.py
```diff
@@ -0,0 +1,120 @@
...
+
+    # category must be one of Active / Away / Unavailable
+    def test_invalid_category_rejected(self):
+        with self.assertRaises(frappe.ValidationError):
+            self._make_status("Busy", category="Busy")
+
+    # only one status may use the Active category (one is already seeded on install)
+    def test_second_active_status_rejected(self):
+        with self.assertRaises(frappe.ValidationError):
+            self._make_status("Online", category="Active")
```
RitvikSardana: why so?

helpdesk/patches/set_default_agent_statuses.py
```diff
@@ -0,0 +1,30 @@
...
+        or_filters=[
+            ["availability", "is", "not set"],
+            ["availability", "=", ""],
+            ["availability", "not in", valid_statuses],
+        ],
+        pluck="name",
+    )
+    for agent in agents:
+        frappe.db.set_value(
+            "HD Agent", agent, "availability", active_status, update_modified=False
```
RitvikSardana: you can directly do bulk_edit using 

frappe.db.set_value("HD Agent", {}, "status", "Active",update_modified=False)

helpdesk/setup/install.py
```diff
@@ -357,6 +358,29 @@ def add_default_status():
...
+            "color": "Green",
+            "order": 1,
+        },
+        {"agent_status": "Away", "category": "Away", "color": "Amber", "order": 2},
+        {
+            "agent_status": "Unavailable",
+            "category": "Unavailable",
+            "color": "Red",
+            "order": 3,
+        },
```
RitvikSardana, line 380: let us do an order 

10 
20
30


because if someone wants to add another status
they can easily insert in between these


if it is


1
2
3


then if user wants to add a status and move it to 3

then the user will have to first update the order of 3rd status to the 4th and then create a 3rd one.


also when I worked on 

HD Agent Status => I used "order" field which is a reserved keyword in MariaDB, so always I had to do

``tabHD Ticket Status`.order"`

so it would have been better to use the name

`status_order` would be better to use in this case also

RitvikSardana (dismissed):

PR #3387, opened 2026-05-25
refactor: add Oauth support in Helpdesk and Simplify Email Accounts connection

aerodeval:
### Current Oauth implementation

Currently to setup Oauth in frappe based apps the flow is

```
  =>  Create a connected app and connect it to Provider

         =>  Create a new Email account select oauth option and use the connected app & user you have made

                => Authorise the API 
```
docs: https://docs.frappe.io/framework/user/en/guides/app-development/connected-app

this process is entirely done on desk and is very confusing for new users who are unfamiliar with the whole ordeal. To tackle this issue ideally Oauth implementation should be made better from Frontend and user should have an easy to use UX.

### Approach

1. Use Connected App Templates for clients that offer Oauth
2. Hide all the Frappe-specific plumbing (Connected App creation, scope wiring, redirect URI construction) behind simple calls in the frontend so that the user never touches a DocType directly.
3. One connected app per provider per site the redirect URI thus is stable and only needs to be registered once regardless of how many mailboxes use it.


aerodeval:
closed for now reopening when will work on it again

PR #3393, opened 2026-05-26
refactor: get_users function to optimize role fetching and user retrieval

### LINKED ISSUE #3335: helpdesk pulls all the users in agent's view on initial load
### Information about bug

When agent's view is loaded `get_users` API is called, which loads all the users basic data into pinia store for faster retrieval.
Now assuming you can have more then 100k or upto 300k users, the API response time is way over 1mins, which in turn get cancelled by the CF level rules. It also blocks the listview of tickets until the users are loaded.
ref: https://github.com/frappe/helpdesk/blob/develop/helpdesk/api/session.py#L8

Possible solution:
- Limit the number of users which are loaded initially from API.
- Add function in the store to validate and fetch the missing user data on the fly.
- Store data in pinia store and return.

### Version

frappe: 16.14.0
helpdesk: 1.22.1

### Installation method

None

```shell

```

saadchaudharry:
### Problem

`get_users()` in `helpdesk.api.session.get_users` has an N+1 query pattern on the initial load of the agent view. For every user in the system, `frappe.get_roles(user.name)` is called inside a loop just to derive a role label (Manager / Agent / Guest).

On a system with N enabled users, this results in **1 + N queries** per request.

### Fix

Fetch all agent-related role mappings from `tabHas Role` in a single query, build an in-memory `roles_map` with priority (Manager > Agent > Guest), then assign labels during the user iteration via dict lookup.

### Impact

| Metric | Before | After |
|---|---|---|
| Queries | 1 + N | 2 |
| Role resolution | DB call per user | In-memory dict lookup |

### Notes

- No change to API signature or return shape — purely internal refactor.
- Role priority preserved (Manager > Agent > Guest) to match existing behavior.
- `session_user` flag behavior unchanged.
- Removed `distinct=True` from the User query since filtering on the primary table doesn't require it.

### DIFF
--- a/helpdesk/api/session.py
+++ b/helpdesk/api/session.py
@@ -6,22 +6,40 @@ from helpdesk.utils import agent_only
 @frappe.whitelist()
 @agent_only
 def get_users():
-    users = frappe.qb.get_query(
+    session_user = frappe.session.user
+
+    # Fetch agent-related roles in a single query
+    role_rows = frappe.db.get_all(
+        "Has Role",
+        filters={
+            "parenttype": "User",
+            "role": ["in", ["Agent Manager", "Agent", "Guest"]],
+        },
+        fields=["parent", "role"],
+    )
+
+    # Role priority: Manager > Agent > Guest
+    priority = {"Agent Manager": 3, "Agent": 2, "Guest": 1}
+    label = {"Agent Manager": "Manager", "Agent": "Agent", "Guest": "Guest"}
+    roles_map = {}
+    for row in role_rows:
+        if priority.get(row.role, 0) > priority.get(roles_map.get(row.parent), 0):
+            roles_map[row.parent] = row.role
+
+    # Fetch all active users
+    users = frappe.db.get_all(
         "User",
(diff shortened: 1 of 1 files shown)


RitvikSardana:
@saadchaudhaary 

Linter are failing, please fix the commit message to resolve the linters.

You can update it to: 
refactor: get_users function to optimize role fetching and user retrieval

saadindictrans:
Hi @RitvikSardana 
The commit message is updated to:
refactor: get_users function to optimize role fetching and user retrieval

RitvikSardana:
The linters are failing again, you need to install pre-commit in your local dev

` pre-commit install`


Install and commit again, the failed CI suggests what changes are to be made

saadindictrans:
> The linters are failing again, you need to install pre-commit in your local dev
> 
> ` pre-commit install`
> 
> Install and commit again, the failed CI suggests what changes are to be made

hi @RitvikSardana 
pre-commit checks are passing locally (black, flake8, isort all green).  — Could you please re-review and merge?

image
image

PR #3424, opened 2026-06-05
feat: categories and language for saved replies

ankitpatil-ap:
Hi team

As a helpdesk grows, the saved-reply list gets long fast — and today there's no
way to organise it. This small PR adds two optional fields to HD Saved Reply to
help:

Category — a Link to a new lightweight "HD Saved Reply Category" doctype
   (title, colour, description). Group replies into buckets like Refunds,
   Onboarding, or Billing so agents find the right one quickly.

Language — a Link to Frappe's built-in Language doctype, so multilingual
   teams can keep per-language variants of the same reply.

Both fields are completely optional and fully declarative — no patch needed,
bench migrate creates the new doctype + columns. Permissions follow the existing
HD Saved Reply roles (System Manager / Agent Manager manage; Agent reads).

This PR is just the data model — the saved-reply picker could later filter by
these. Totally open to your preference on naming or approach (e.g. tags instead
of a category doctype). Happy to iterate! 🙌

Thanks for the great project!

### DIFF
--- a/helpdesk/helpdesk/doctype/hd_saved_reply/hd_saved_reply.json
+++ b/helpdesk/helpdesk/doctype/hd_saved_reply/hd_saved_reply.json
@@ -8,6 +8,8 @@
  "engine": "InnoDB",
  "field_order": [
   "title",
+  "category",
+  "language",
   "scope",
   "teams",
   "section_break_ybsb",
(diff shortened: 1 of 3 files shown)


RitvikSardana:
Instead of `category`, we can categorise it team-wise

RitvikSardana:
Going ahead with team wise only, hence closing

PR #3425, opened 2026-06-05
feat: auto-assign ticket type from keywords

ankitpatil-ap:
Hi team

A small quality-of-life addition for teams that triage by ticket type. Each
HD Ticket Type can carry a list of keywords, and when a new ticket arrives
without a type, it's auto-assigned to the type whose keyword matches its
subject/description.

How it works:
    New optional "Keywords" field on HD Ticket Type (comma-separated, e.g.
   "refund, payment, invoice").
    On a new ticket with no type, the subject + description are matched against
   those keywords. Matching is word-boundary aware ("pay" won't match "happy"),
   the longest matching keyword wins, and it only runs when the type is empty 
   so it never overrides an explicit choice or the default type.
   The keyword map is cached and invalidated when a ticket type changes.

Fully opt-in no keywords means no behaviour change.

I know this overlaps conceptually with Assignment Rule. I went with a tiny
keyword field on the type because it's simple and self-contained, but I'm very
happy to rework it as an Assignment-Rule-based approach if you'd prefer that
direction. Let me know what fits best! 

Thanks for the great project!

### DIFF
--- a/helpdesk/helpdesk/doctype/hd_ticket/hd_ticket.py
+++ b/helpdesk/helpdesk/doctype/hd_ticket/hd_ticket.py
@@ -1,4 +1,5 @@
 import json
+import re
 import uuid
 from datetime import timedelta
 from email.utils import parseaddr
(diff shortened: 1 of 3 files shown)


RitvikSardana:
This is not required, can be part of automations feature in the future

PR #3488, opened 2026-06-20
fix: dedupe recent tickets, feedback rating by resolution date, staledashboard stats

RitvikSardana:
- get_recent_tickets: exclude current + org tickets from the raised_by query so a ticket matching both customer and raised_by no longer appears twice.
- Dashboard avg feedback rating: filter on resolution_date instead of creation, so every ticket resolved in the period is counted regardless of when it was raised.
- Dashboard.vue: drop the static frappe-ui cache keys on the stat resources; the shared cached instance stayed bound to a previous mount's filters and sent stale params after navigating away and back.


RitvikSardana:
@greptileai rereview

PR #3512, opened 2026-06-29
Revert "fix: bump frappe UI"

RitvikSardana:
Reverts frappe/helpdesk#3502


shahzeelahmed:
@RitvikSardana what was the reason for reverting? 🤔

PR #3544, opened 2026-07-13
fix!: field-level permissions for agent-only ticket fields

aerodeval:
## Problem

From the portal a customer could set agent-only fields on their own ticket — `priority`, `sla`, `response_by` rest api methods

## Fix

- **Reading** — permission levels on HD Ticket: **7** customer-readable, **8** agent-only. Frappe enforces them on every path, `/api/resource` included. 1–6 stay free for a site's own scheme.
- **Editing** — `prevent_customer_edits()` in `validate`. Levels can't cover this: a customer needs level-0 write just to create a ticket. They keep `feedback`, `feedback_extra` and closing the ticket.
- **Per-site hiding** — each Default-template row carries a **Visible to** select (Everyone / Customers / Agents) that hides the field on helpdesk's own pages. It only narrows what the level already allows, never widens.

| Level | Fields | Customers | Agents |
| --- | --- | --- | --- |
| **0** | `subject`, `status`, `description`, `template`, `feedback`, `feedback_extra`, `feedback_rating` | fill on create, then only status (to close) and feedback | edit |
| **7** | `raised_by`, `contact`, `customer`, `priority`, `ticket_type`, `agent_group`, `sla`, `status_category`, `response_by`, `resolution_by`, `on_hold_since`, `first_response_time`, `first_responded_on`, `first_response_failed_by`, `resolution_date`, `resolution_time`, `resolution_failed_by` | read; fill on create if the template shows it | edit |
| **8** | `key`, `summary`, `resolution_details`, `agreement_status`, `service_level_agreement_creation`, `opening_date`, `opening_time`, `avg_response_time`, `total_hold_time`, `user_resolution_time`, `raised_outside_working_hours`, `last_agent_response`, `last_customer_response`, `email_account`, `via_customer_portal`, `is_merged`, `merged_with`, `ticket_split_from` | no access | edit |

`attachment` and `content_type` — hidden, unused `Issue` leftovers — sit at level 0 by default and are left out above; removing them is a separate change.

Also fixes the fallout of hiding fields: portal reply no longer reopening a ticket, the closing-email rating link, 500 on tickets with no contact, merge losing `is_merged`, blank list columns and filters that raised `PermissionError`, Default template rename, and an agent's whole-document save blanking hidden fields.

## Breaking changes

- Customers can no longer edit ticket fields after creation (closing and feedback still work).
- `hide_from_customer` is replaced by **Visible to**; a patch carries old values over. Update anything reading the flag.
- Hiding a template field hides it on helpdesk pages only — raise its level in Customize Form to hide it from the REST API too.
- A non-agent creating a ticket — including a System Manager with no HD Agent record — now gets `raised_by` forced to their own address and `via_customer_portal` set. System Manager keeps full field access through the new level-7/8 perm rows.

## Existing sites

`harden_hd_ticket_field_permissions` writes the level-7/8 perm rows (a single `Custom DocPerm` row makes Frappe ignore shipped perms, so a customised site would read every protected field as empty — agents too). `backfill_template_field_visible_to` moves old `hide_from_customer` values across. Both idempotent, neither writes a field's level.

## Testing

118 tests in `test_hd_ticket.py`, 20 in `test_field_visibility.py`, 2 in `test_hd_email_feedback.py`.


desk/src/components/ticket/TicketCustomerSidebar.vue
```diff
@@ -201,8 +185,9 @@ const ticketAdditionalInfo = computed(() => {
   const custom_fields = ticket.data.template.fields
     .filter(
       (field: Field) =>
-        !field.hide_from_customer &&
-        ["subject", "team", "priority"].indexOf(field.fieldname) === -1
+        field.visible_to !== "Agents" &&
```
ps173: hard coded to not show to agents ?

desk/src/pages/ticket/TicketCustomerTemplateFields.vue
```diff
@@ -89,7 +89,7 @@ const slaData = computed(() =>
 
 const customFields = computed(() => {
   const _custom_fields = ticket.data.template.fields
-    .filter((field: Field) => !field.hide_from_customer)
+    .filter((field: Field) => field.visible_to !== "Agents")
```
ps173: The server already drops agent-only rows in get_fields_meta via hidden_ticket_fields(), so this filter (and the same ones in TicketCustomerSidebar and TicketCustomerTemplateFields) never removes anything for a customer. Can we drop the three client-side checks and the visible_to key on Field, so visibility has one owner on the server?

desk/src/types.ts
```diff
@@ -247,7 +247,7 @@ export interface AutoCompleteItem {
 export interface Field {
   fieldname: string;
   fieldtype: string;
-  hide_from_customer?: 0 | 1;
+  visible_to?: "Everyone" | "Customers" | "Agents";
```
ps173: should be rather based on roles ? This don't make a lot of sense to have hard coded strings. and does agents comprise of agent managers ?

desk/src/components/ticket/TicketCustomerSidebar.vue
aerodeval: yes since its customer sidebar component (they get fields visible_to which are set as customers and everyone visibility)

helpdesk/api/doc.py
```diff
@@ -389,6 +403,9 @@ def get_filterable_fields(
     for field in standard_fields:
         if field.get("fieldname") not in [r.get("fieldname") for r in res]:
             res.append(field)
+    if doctype == "HD Ticket":
```
ps173, line 406: This `if doctype == "HD Ticket"` + `hidden_ticket_fields()` + filter block is now repeated in `get_list_data`, `get_filterable_fields`, `sort_options` and `get_quick_filters`. Could we fold the doctype check into a single helper so each call site is one line?

```python
def hidden_fields(doctype: str) -> set[str]:
    hidden = unreadable_fields(doctype)
    if doctype == "HD Ticket":
        hidden |= ticket_visibility_tier()
    return hidden
```

Then here it is just:

```python
res = [f for f in res if f["fieldname"] not in hidden_fields(doctype)]
```

helpdesk/helpdesk/doctype/hd_ticket/hd_ticket.py
```diff
@@ -71,9 +79,50 @@ def autoname(self):
...
+        if is_agent():
+            return
+        if frappe.session.user != "Guest":
+            self.raised_by = frappe.session.user
+        self.via_customer_portal = 1
+        self.flags.ignore_permlevel_for_fields = (
+            list(PORTAL_INSERT_EXEMPT_FIELDS) + self.creation_fillable_template_fields()
+        )
+
+    def creation_fillable_template_fields(self) -> list[str]:
```
ps173: rename to `customer_fillable_template_fields` ( matches context )

desk/src/pages/ticket/TicketNew.vue
```diff
@@ -234,7 +234,7 @@ const customOnChange = computed(() => template.data?._customOnChange);
 
 const visibleFields = computed(() => {
   let _fields = template.data?.fields?.filter(
-    (f) => !isCustomerPortal.value || !f.hide_from_customer
+    (f) => !isCustomerPortal.value || f.visible_to !== "Agents"
```
ps173: same as last one, this is a redundant check

helpdesk/helpdesk/doctype/hd_ticket_template/hd_ticket_template.py
```diff
@@ -43,6 +52,62 @@ def validate_unallowed_fields(self):
...
+                "{0} are hidden from customers here, but the API still returns them."
+                " Raise their permission levels in {1} to hide them everywhere."
+            ).format(comma_and(exposed, add_quotes=False), link)
+        frappe.msgprint(text, title=_("Information"), indicator="blue")
+
+    def current_permlevel(self, fieldname: str) -> int:
+        """Live meta, so a level changed in Customize Form counts."""
+        field = frappe.get_meta("HD Ticket").get_field(fieldname)
+        return field.permlevel if field else 0
+
```
ps173, line 135: uhh what

aerodeval: so lets say someone changes the fields permlevel using customize form  

for example
agent_group permlevel 7 from permlevel 8

and then adds the field in ticket template, it wouldn't show any warning because as per permlevel user is supposed to see this should be a on the fly check what the perm level is

helpdesk/api/doc.py
aerodeval: each has seperate context as get_list_data is not just for Hd Ticket doctype but also for overall other lists, permlevel and meta filtering is now just for HD Ticket so only do that when the doctype is HD ticket

helpdesk/helpdesk/doctype/hd_ticket_template/hd_ticket_template.json
```diff
@@ -28,6 +28,7 @@
    "options": "HTML"
   },
   {
+   "description": "Visible To controls the visibility of the field. Setting it to Customers will allow only customers to be able to see that field, Agents only agents, whereas if set to Everyone both agents and customers can see and fill it while raising a ticket. To know more read more about the documentation <a href=\"https://docs.frappe.io/framework/user/en/basics/users-and-permissions\" target=\"_blank\" rel=\"noopener noreferrer\">here</a>.",
```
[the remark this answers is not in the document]
aerodeval: this is as expected because while adding a field in template you are given a warning about permlevels being restricted

helpdesk/helpdesk/doctype/hd_ticket_template/api.py
```diff
@@ -1,36 +1,32 @@
...
-    ) or [None, None, None]
-    if not found:
+    template = frappe.db.get_value(
+        DOCTYPE_TEMPLATE, name, ["about", "description_template"], as_dict=True
+    )
+    if not template:
         return {"about": None, "fields": []}
-
-    fields = get_fields_meta(name)
+    fields = TicketFields().form
```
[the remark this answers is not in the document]
aerodeval: all logic is built on Default Template. There is only one template as of now that is the approach we are following

helpdesk/helpdesk/doctype/hd_ticket/hd_ticket.py
```diff
@@ -167,6 +207,8 @@ def handle_email_feedback(self):
             frappe.throw(_("Could not send feedback email,due to: {0}").format(e))
 
     def after_insert(self):
+        # the creation-form exemption must not survive into a later save
```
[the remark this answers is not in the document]
aerodeval: special exemption for this time as we need to ignore permlevels for fields at first insert and avoid later. No native fw way to do this

PR #3546, opened 2026-07-13
fix(search): avoid SQLite variable limit in filter facets

### LINKED ISSUE #3250: sqlite search throws too many variables error
### Information about bug

Core  ref issue : https://github.com/frappe/frappe/issues/38939

HelpdeskSearch class  has method called `get_filter_options` which filters the results based on the allowed tickets and same as core sqlite query  filter options breaks because of too many variables.

### Version

v1.22.1

### Installation method

None

```shell

```

Dkm0315:
## What this fixes

Helpdesk's SQLite facet query repeated the accessible ticket list in three `IN (...)` clauses (`name`, `reference_name`, and `reference_ticket`). For `N` accessible tickets, that generated `3N` host parameters. A 334-ticket access list therefore produced 1,002 binds and failed when SQLite's variable limit was 999.

This change serializes the access list once, expands it with `json_each(?)` in an `accessible_tickets(ticket)` CTE, and keeps the three access checks as subqueries against that CTE. The query now uses one bound parameter, retains the existing facet semantics, and remains `read_only=True`.

Fixes #3250

### Bug — unchanged query

With 334 accessible ticket IDs and SQLite limited to 999 variables, the facet request fails with `sqlite3.OperationalError: too many SQL variables`.

Bug: Helpdesk facet request fails with too many SQL variables

### Fix — same boundary

After the change, the authenticated Helpdesk search page loads its Status facets under the same forced limit.

Fix: Helpdesk Status facets load at the same SQLite boundary

The fixed live request returned HTTP 200 with:

```json
{"teams": {}, "statuses": {"Open": 334}, "priorities": {"Medium": 334}, "customers": {}, "doctypes": {"HD Ticket": 334}}
```

Both captures used the same 334 controlled Open tickets and the same forced limit. The temporary tickets and forced-limit instrumentation were removed after capture and are not included in this branch.

## Verification

| Check | Result |
|---|---|
| Focused SQLite search regression suite | 4/4 passed |
| Full Helpdesk suite | 248/248 passed (134 integration + 114 old-Frappe category) |
| Mutation/red proof | Old query fails at 334 IDs / limit 999 with `too many SQL variables` |
| 10,000-ID scale assertion | One bind; `read_only=True` |
| Targeted pre-commit | Black, strict flake8, and isort passed |
| `bench setup requirements` | Passed |
| `bench build` | Passed, including Helpdesk assets |
| `bench --site high-reach.localhost migrate` | Passed |
| Authenticated facet API and search UI | HTTP 200; Status facets rendered |

The regression coverage includes:

- exact Ticket, Comment, and Communication facets at the 334-ID boundary;
- inaccessible results excluded from every facet;
- duplicate IDs, Unicode, and quoted names;
- empty access lists and a missing search index;
- no double count when a result matches more than one reference column;
- 10,000 IDs with one bind and read-only execution.

## Query-size and timing check

Representative local timings for the exact queries against 15,000 indexed rows:

| Accessible IDs | Previous query | JSON/CTE query |
|---:|---:|---:|
| 1,000 | 13.28 ms | 13.98 ms |
| 10,000 | 33.47 ms | 29.16 ms |
| 50,000 | 54.83 ms | 46.89 ms |

At 50,000 IDs, the SQL statement and bind count drop from 300,205 characters / 150,000 binds to 387 characters / 1 bind. The JSON payload itself remains linear in the number of accessible tickets.

## Scope

This PR fixes Helpdesk's SQLite facet-filter path only. [`frappe/frappe#38939`](https://github.com/frappe/frappe/issues/38939) tracks the separate Frappe core search path; this PR does not claim to fix that companion issue.

@RitvikSardana, could you please review this and, if appropriate, assign #3250 to @Dkm0315?

### DIFF
--- a/helpdesk/search_sqlite.py
+++ b/helpdesk/search_sqlite.py
@@ -131,8 +131,16 @@ class HelpdeskSearch(SQLiteSearch):
                 "doctypes": {},
             }
 
-        # Query the search index for available options
+        accessible_tickets_json = frappe.as_json(
+            accessible_tickets, indent=None, separators=(",", ":")
+        )
+
+        # Query the search index for available options. Binding the accessible
+        # tickets as one JSON value avoids SQLite's statement-wide variable limit.
         sql = """
+			WITH accessible_tickets(ticket) AS (
+				SELECT value FROM json_each(?)
+			)
 			SELECT
 				agent_group,
 				status,
@@ -141,14 +149,15 @@ class HelpdeskSearch(SQLiteSearch):
 				doctype,
 				COUNT(*) as count
 			FROM search_fts
-			WHERE (name IN ({placeholders}) OR reference_name IN ({placeholders}) OR reference_ticket IN ({placeholders}))
+			WHERE (
+				name IN (SELECT ticket FROM accessible_tickets)
+				OR reference_name IN (SELECT ticket FROM accessible_tickets)
+				OR reference_ticket IN (SELECT ticket FROM accessible_tickets)
+			)
 			GROUP BY agent_group, status, priority, customer, doctype
-		""".format(
-            placeholders=",".join(["?" for _ in accessible_tickets])
-        )
+		"""
 
-        params = accessible_tickets * 3
-        results = self.sql(sql, params, read_only=True)
+        results = self.sql(sql, (accessible_tickets_json,), read_only=True)
 
         # Aggregate the results
         teams = {}
--- a/helpdesk/test_search_sqlite.py
+++ b/helpdesk/test_search_sqlite.py
@@ -1,0 +1,171 @@
+import sqlite3
+from unittest.mock import patch
+
+import frappe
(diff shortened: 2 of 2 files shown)


Dkm0315:
## Actual bug and fix evidence

Both captures use the identical fixture: **334 accessible Open tickets** with SQLite's variable limit forced to **999**. Temporary data and instrumentation were removed afterward.

### Bug — unchanged query

The unchanged 3N query creates 1,002 binds and raises **sqlite3.OperationalError: too many SQL variables**.

clipboard


### Fix — same boundary

The fixed query binds the access list once. The same request returns HTTP 200 with Open: 334, and the authenticated Helpdesk UI loads the Open Status facet.

clipboard


Fixed response: **Open 334**, **Medium 334**, **HD Ticket 334**. Focused tests: **4/4**; full Helpdesk suite: **248/248**.

RitvikSardana:
This has been fixed

PR #3547, opened 2026-07-14
fix: ticket SLA policy application

aerodeval:
## Summary
- An enabled, non-default SLA with a blank `condition` was matching every ticket ahead of the actual default SLA, because a blank condition short-circuited to "always match" in `get_sla()`'s loop.
- Most visible when the seeded "Default" SLA is demoted (a different SLA gets `default_sla` checked) but keeps its blank condition — it then silently wins for every ticket regardless of which SLA is actually flagged as default.
- Fix: exclude SLAs with a blank/null condition from the conditional-match query, so `default_sla` is the only legitimate catch-all.

## Test plan
- [x] Create SLA "Default" with `default_sla=1`, no condition (matches install fixture)
- [x] Flag a second SLA as default (`default_sla=1`); confirm "Default" is demoted and its condition remains blank
- [x] Create a ticket and confirm it resolves to the newly-flagged default SLA, not "Default"
- [x] Existing `test_hd_service_level_agreement.py` / `test_hd_ticket.py` SLA tests still pass

### DIFF
--- a/desk/src/components/Settings/Sla/SlaPolicyView.vue
+++ b/desk/src/components/Settings/Sla/SlaPolicyView.vue
@@ -481,9 +481,9 @@ const toggleEnabled = () => {
   slaData.value.enabled = !slaData.value.enabled;
 };
 
-const toggleDefaultSla = () => {
-  slaData.value.default_sla = !slaData.value.default_sla;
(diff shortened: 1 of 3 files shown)


RitvikSardana:
@aerodeval add a test case too?

Issue #3518, opened 2026-07-02
Bug: "Created On" date filter returns no results when using the "Between" operator.


omisocial:
### **Issue Detailed Description:**

The "Created On" date filter is malfunctioning when users select the **"Between"** condition. Even though valid data exists within the selected date range, the system incorrectly displays an empty state (no results found).

A screen recording and console log have been captured, indicating a potential front-end state issue or API payload mismatch when handling the date range parameters.

---

### **Steps to Replicate the issue:**

1. Navigate to the main dashboard/list view where the data table is located.
2. Click on the filter options and select the **"Created On"** date field.
3. Choose **"Between"** as the filter operator.
4. Select a valid start date and end date range that is known to contain existing records.
5. Click "Apply" or trigger the filter search.

---

### **Current behavior:**

The table returns zero results ("No data available"), despite records existing within that specific timeframe.

---

### **Expected behavior:**

The table should correctly filter and display all records that were created within the selected "Between" date range.

<img width="1405" height="733" alt="Image" src="https://github.com/user-attachments/assets/070e39b6-2acb-452e-990d-ce3ef54e2f16" />

### Version

<img width="1289" height="279" alt="Image" src="https://github.com/user-attachments/assets/a3db725c-c630-4523-91eb-582f9226e2bf" />

### Installation method

None

### Relevant log output / Stack trace / Full Error Message.

```shell
### **Supporting Screenshots/documents/videos/dashboard links:**

* **Jam Screen Recording & Console Logs:** [https://jam.dev/c/39308e06-3c13-4c07-9461-14c0f21eb83f](https://jam.dev/c/39308e06-3c13-4c07-9461-14c0f21eb83f)
```

RitvikSardana:
This was fixed by the filter rewrite (#3447, #3475). The Created On "Between" filter now passes a [from, to] range and the end date is expanded to the end of the day. Please reopen with your Helpdesk version if you still see it on the latest release.

PR #3550, opened 2026-07-14
fix: correct Created On date range filter behavior

harrishragavan:
# Fix: "Created On" `Between` Date Filter Returns No Results (#3518)

## Problem

The **Created On** filter does not work correctly when the **Between** operator is selected. Even when records exist within the selected date range, the ticket list incorrectly displays **"No data available."**

This issue occurs because the front-end does not consistently handle the value emitted by `DateRangePicker` when constructing the filter value.

## Root Cause

`DateRangePicker` emits the selected date range as a `string[]` (`[startDate, endDate]`), while parts of the filtering logic expected a comma-separated string.

As a result:

* The **Between** filter value was not parsed consistently.
* Date range values were processed incorrectly before the filter was applied.
* An invalid filter value was generated, causing records that fall within the selected date range to be excluded from the results.

## Solution

The date range handling has been updated to support both array and string inputs, ensuring the **Between** filter is parsed correctly regardless of the value format returned by `DateRangePicker`.

The implementation now:

* Accepts the `string[]` value emitted by `DateRangePicker`.
* Maintains compatibility with legacy comma-separated string values.
* Safely handles unexpected or empty inputs.
* Generates the correct filter value so records within the selected date range are returned as expected.

This restores the expected behavior for the **Created On → Between** filter while preserving compatibility with existing filter functionality.
image


harrishragavan:
@RitvikSardana Kindly review this PR.

PR #3554, opened 2026-07-16
added missing msgstr for translation

ews-pgasser:


aerodeval:
@ews-pgasser  good to merge?

PR #3587, opened 2026-07-19
ci: parallelize server tests 4-way and speed up the install phase

RitvikSardana:
## What

Cut Server CI wall-clock (was ~8m40) with two independent optimizations, plus a codecov fix.

### 1. Parallelize the test phase (4-way)
`Python Unit Tests` is now a 4-job matrix using Frappe's built-in split runner (`run-parallel-tests --total-builds 4 --build-number N`). ERPNext stays installed and its integration tests run on **every** PR as part of the normal suite. No test isolation, no path-gating.

### 2. Speed up the install phase
The install phase is now the floor, so this ports the modern `frappe/frappe` setup-action techniques:

- **MariaDB in tmpfs** (data dir in RAM) so site reinstall and the DB-bound test setup stop hitting disk.
- **Durability off** on the ephemeral CI DB (`innodb_flush_log_at_trx_commit=0`, `sync_binlog=0`) to skip per-commit fsyncs.
- **`uv tool install frappe-bench`** (with cached `setup-uv`) instead of `pip`.
- **Overlap independent steps**: apt system deps install while the bench tool installs; the ERPNext clone runs during `bench init`.
- **Shallow-clone ERPNext** (`--depth 1`) instead of a full-history clone.

### 3. Codecov
Added `codecov.yml` (helpdesk had none, so strict defaults failed the project status on any coverage drop). Mirrors frappe/erpnext: project status tolerates a 1% dip; new code gated at 85% patch coverage (PR-only, ignored when tests fail).

## Note for required checks
The check name changes from `Python Unit Tests` to `Python Unit Tests (1)` … `(4)`. Branch protection / mergify rules referencing the old name need updating.

## How to test
CI on this PR: the four matrix jobs should pass with roughly a quarter of the tests each, ERPNext present in the install log, and a faster install phase.


RitvikSardana:
@greptileai rereview

RitvikSardana:
@greptileai rereview

PR #3623, opened 2026-07-27
fix(dashboard): show chart-specific empty states

RitvikSardana:
## Problem

Every empty chart on the dashboard showed the same generic line: *"Dashboard charts will appear here once you start receiving or creating tickets."*

All three call sites (trend charts, master-data charts, tag charts) passed only a `title` override to `SkeletonLoader`, so `message` always fell through to the default. The empty cards also rendered grey skeleton bars where the chart header should be, which looked broken sitting next to real charts.

## What this does

- `Dashboard.vue`: single `emptyStateByChart` map keyed by chart title, with specific copy per chart. `chartEmptyState(chart)` replaces the three duplicated inline objects, and the all-empty dashboard state derives its list from the same map.
- `SkeletonLoader.vue`: renders the real chart title and subtitle when passed, grey placeholder bars otherwise.

Unknown chart titles still fall back to `No <title> available`.

## How to test

1. Open the dashboard on a site with tickets but no tags.
2. The Top Tags and Tag Trend cards should show their real headers plus "No tags used yet. / The most used tags will be ranked here once tickets are tagged." instead of the generic dashboard message.
3. Apply a date range with no tickets. Every chart card should show its own title and the filter message.

### DIFF
--- a/desk/src/components/SkeletonLoader.vue
+++ b/desk/src/components/SkeletonLoader.vue
@@ -54,8 +54,18 @@
           :key="`chart-${i}`"
           class="border rounded-md min-h-[300px] p-4 space-y-3"
         >
-          <div class="h-4 w-1/3 bg-surface-gray-3 rounded" />
-          <div class="h-4 w-1/4 bg-surface-gray-1 rounded" />
+          <template v-if="getEmptyState(i).chartTitle">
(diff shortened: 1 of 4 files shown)


RitvikSardana:
Only the Docker one is the real fix; other than that, all of it seems irrelevant
Please fix that in a separate PR; closing this

Issue #1213, opened 2023-05-22
Cannot reply to ticket


michellealva:
<img width="1665" alt="image" src="https://github.com/frappe/helpdesk/assets/50285544/b6b0fa2e-403a-4da3-b4fa-0409fb03f9ba">

P.s. comments work though

prasathsekar:
![Screenshot 2023-05-26 at 5 48 18 PM](https://github.com/frappe/helpdesk/assets/25291469/29f21cd2-04c4-4f7d-93e1-ba62b1d4f01e)
Getting Same Error, unable to reply by tagging the email.

Comment and Reply is working.

ssiyad:
This is old version of Helpdesk. Update to latest and try

prasathsekar:
How can i find the current version and how can i update it to the new version 

Any material guidance is there?

ssiyad:
@prasathsekar Which commit? Even though there are versioned releases, you can always jump ahead to newest commits. Also, the screenshot you see in the beginning is old design. Hence, is an old version

prasathsekar:
![Screenshot 2023-05-26 at 7 22 33 PM](https://github.com/frappe/helpdesk/assets/25291469/54d1f321-333d-4e03-a4e7-2ae9074e2358)
Screenshot for your referance

ssiyad:
In that case, create a different ticket. This seems unrelated to original issue

PR #3626, opened 2026-07-27
feat(command-palette): contextual Cmd+K palette for the agent portal

RitvikSardana:
## What this does

Rebuilds the agent portal's `Cmd+K` palette around **what you're currently looking at**, instead of a flat list of links.

**Ticket context** (on a ticket) — change status, priority, assignee, team, type; add a tag; comment; reply; **insert a saved reply**; merge; next/previous ticket; copy id/link. Sub-lists load their options on demand. Typing an option name directly (`urgent`, `open`) surfaces a flat row that sets it in one Enter, no drill-down.

**Ticket list context** (on the list) — filter by status or priority, show tickets assigned to me, switch view, clear filters. Filters merge through the list's own `applyFilters`, so applying one **preserves your sort, columns and active view**.

**Everywhere** — server-side FTS search over tickets, comments and communications merged into the same ranked list as commands, with the server's `<mark>` highlighting rendered as inert text runs; recently-visited tickets on an empty query; a Settings drill-down derived from the permission-filtered `tabs` list, so there's no second copy of who-can-see-what.

**Saved replies** — `Cmd+K → type → Enter` opens the composer pre-filled with the rendered reply (never send-on-select). A drill-down lists all replies ranked by per-agent usage; an agent's top 3 most-used replies surface as root-level rows once they cross a usage threshold.

Rows show the app's existing shortcut where one exists (`S`, `P`, `A`, `R`, `C`, `G`, `⌘,` …), matching what `ShortcutsModal` already advertises — the palette teaches the keyboard rather than replacing it.

Scope **ranks, it never gates**: on a ticket, that ticket's actions rank above search hits and navigation, but nothing is excluded — a ticket whose subject contains "open" stays reachable even though `Set status: Open` matches too.

Accessibility and keyboard: combobox/listbox/option roles with a live results region, keys bound to the dialog rather than the input, `Home`/`End` and `Ctrl+N`/`Ctrl+P`, focus restored to where it was on close, stale search responses dropped instead of flashing empty.

Telemetry: `command_palette_opened` and `command_palette_command_run` (command id, query length, depth), plus `source` on `ticket_assigned` and `saved_reply_applied` so palette-driven actions are distinguishable from their popover/modal equivalents.

Architecture and rationale are documented in `desk/src/components/command-palette/README.md`; remaining ideas are tracked in `CP-improvements-plan.md`.

## Supporting changes

- **`translation.ts`** — `__()` silently dropped every replacement after the first when given an array; it now accepts arrays, fixing 47 existing call sites app-wide and cutting the repo's `tsc` error count from 952 to 709.
- **`ListViewBuilder.vue`** — `handleViewChanges()` did two jobs, and reassigned `order_by`/`columns`/`rows` from the view before applying URL filters. Split into `switchToView()` (owns sort/columns/rows, runs only when `?view` changes) and `applyUrlFilters()` (touches filters only). Fixes filters silently resetting the user's sort. An unresolvable `?view` now redirects while keeping the filters instead of dropping them.
- **`listViewFilters.ts` / `replyComposer.ts`** — the palette mounts in the sidebar, outside `ListViewBuilder`'s and `EmailEditor`'s `provide()` chains, so each publishes a module-scope `shallowRef` the mounted component fills and clears; the palette handles their absence.
- **`useTicketNavigation.ts`** — navigation functions hoisted to module scope so the palette can reuse them; the composable is now shortcut registration only. Its return value was unused.
- **`modalStates.ts`** — merge modal state shared (as the reply/comment boxes already are) so the palette can open it.
- **`layoutSettings.ts`** — navigate commands derive from the sidebar's own nav list, so labels and icons can't drift. Contacts icon unified across both surfaces.
- **`userStorage.ts`** — palette localStorage keys are scoped per session user, so shared machines don't leak one agent's recents to the next.
- **`components/index.ts`** — dropped the `CommandPalette` barrel export. The palette is a singleton registering a global shortcut; a second mount meant two `Cmd+K` handlers on one `isOpen`.
- **`AssignTo.vue`** — fixed a pre-existing malformed `capture()` payload.

## Screenshots

**Ticket context** — the ticket's own actions lead, each printing the shortcut the app already binds. Sub-lists are marked with a chevron.

Command palette on a ticket, showing ticket actions with keyboard shortcuts

**Drill-down** — options load on demand and reuse the app's own icons, so priorities render `TicketPriority.vue` rather than a flat glyph.

Change priority drill-down showing per-level priority icons

**Ticket list context** — filters, views and "assigned to me". Filters merge through the list's own state and no longer reset your sort.

Command palette on the ticket list, showing filter and view commands

**Search** — SQLite FTS hits ranked into the same list, leading with priority and carrying `status · #id` so repeated subjects stay distinguishable.

Search results for login, each row showing a priority icon and status with ticket id

**Settings** — drilled from the permission-filtered `tabs` list, keeping the settings modal's own section headers.

Settings drill-down grouped by My settings, Email Settings and App Settings

## How to test

1. `cd desk && yarn build`, then hard-reload (unregister the service worker).
2. **Ticket page** → `Cmd+K`: ticket actions rank above everything. Drill into "Change priority", pick one — it applies and the palette closes. Type `urgent` at the root instead — `Set priority: Urgent` leads and applies in one Enter.
3. **Saved replies** → `Cmd+K` → "Saved Replies" → pick one: the reply box opens pre-filled with the rendered template. Apply the same reply 5 times and it appears as a root-level row.
4. **Ticket list** → sort by any column, then `Cmd+K` → "Filter by status" → "Open". Filter applies **and the sort is preserved**. "Clear filters" clears filters only.
5. Type `#1213` → jumps straight to that ticket. Type a subject fragment → FTS results with priority icon, `status · #id`, and the matched term highlighted.
6. Empty query → recently-visited tickets (the currently open ticket is excluded).
7. Arrow to the bottom of a long list with the mouse parked mid-list — the highlight must not jump back to the cursor.
8. `Esc` peels one layer at a time: clear query → pop drill-down → close.

## Notes

- No issue to link — this came from a direct request rather than a tracked issue.
- Frontend only. The related backend work — keeping search's permission filtering bounded at any ticket count — shipped separately in #3629 (merged), and `develop` is merged into this branch.
- No test framework in `desk/`, so each pure module carries a runnable assert file: `fuzzyScore.check.ts` (ranking, flat-row vs parent, `<mark>` run splitting), `savedReplyRanking.check.ts` (promotion thresholds, sort stability), `listViewFilters.check.ts` (merge/replace/clear). Run with `npx tsx <file>`. Everything else was verified manually.


RitvikSardana:
@greptileai rereview

PR #3627, opened 2026-07-27
fix(assignment): close related ToDo when a ticket resolves

RitvikSardana:
### Problem

Assignment ToDos are created when an agent is assigned and then never change state. Resolving or closing a ticket leaves its ToDo `Open`, so every ticket an agent has ever handled keeps sitting in their ToDo list.

### Fix

Sync the ToDo status with the ticket's status category on update:

- category `Resolved` (statuses **Resolved**, **Closed**) closes the ToDos
- any other category (`Open`, `Paused`) reopens them, so un-resolving a ticket brings the assignment back

Only fires when `status_category` actually changed, and only flips rows already in the opposite state.

Written via `db.set_value` rather than saving the ToDo: saving drops the agent from `_assign`, and Helpdesk treats `_assign` as the ticket's assignee for its whole lifetime, not just while it is open.

### How to test

1. Assign yourself to a ticket, confirm the ToDo is `Open`.
2. Set the ticket to **Resolved**. The ToDo goes `Closed`, and you are still on `_assign`.
3. Set it back to **Open** or **Replied**. The ToDo goes `Open` again.

Covered by `test_todo_status_follows_ticket_status_category`, which walks Open, Resolved, Replied, Open and asserts `_assign` survives the round trip.

```
bench --site hd-tests run-tests --module helpdesk.helpdesk.doctype.hd_ticket.test_hd_ticket --test test_todo_status_follows_ticket_status_category
✔ test_todo_status_follows_ticket_status_category
```

### DIFF
--- a/helpdesk/helpdesk/doctype/hd_ticket/hd_ticket.py
+++ b/helpdesk/helpdesk/doctype/hd_ticket/hd_ticket.py
@@ -246,6 +246,10 @@ class HDTicket(Document):
     def sync_assignment_todos(self):
         """Close assignment ToDos once the ticket resolves, reopen them if it reopens.
 
+        Scoped to allocated ToDos, which is how `ToDo.update_in_reference` defines
+        an assignment, so ToDos merely referencing the ticket are left alone.
+        Cancelled ToDos stay cancelled, a removed assignment must not come back.
+
         Uses `db.set_value` because saving a ToDo drops the agent from `_assign`,
         which Helpdesk treats as the assignee for the ticket's whole lifetime.
         """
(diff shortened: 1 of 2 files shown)


RitvikSardana:
@greptileai rereview

RitvikSardana:
Closing. Frappe defines an assignment as an open ToDo: `ToDo.update_in_reference` rebuilds `_assign` from ToDos with `status not in (Cancelled, Closed)`, and `assign_to.get` filters the same way. Closing the ToDo on resolve therefore means unassigning, and the `db.set_value` workaround only hides that until the next ToDo write on the ticket recomputes `_assign` and silently drops the agent.

The real ask (agents' ToDo lists filling up with resolved tickets) is a read-side filtering problem, not a ticket-status one.

PR #3635, opened 2026-07-30
refactor: agent availability through the document lifecycle

aerodeval:
## Problem

`set_my_availability` wrote with `frappe.db.set_value`, which skips the HD Agent controller entirely. That put three invariants inside one API function and left every *other* write path to `HD Agent.availability` broken:

1. **Duplicated logic** — `HDAgent.before_save` already stamps `availability_changed_on`, but never ran, so the API re-implemented the stamp by hand.
2. **Validation in the wrong place** — the "status must exist and be enabled" check lived in the API. `availability` is a plain Link to HD Agent Status with no `enable` filter, so the desk form, REST and `frappe.client.set_value` could all write a disabled status silently.
3. **Realtime coupled to one caller** — `agent_availability_updated` was published only by that API, so a status changed from the desk form never reached connected clients.

## Change

Move all three onto the controller, so every write path is correct rather than just the one the endpoint owned:

- `validate()` → rejects a disabled or unknown status. Guarded on `has_value_changed` so an agent already sitting on a since-disabled status can still save unrelated fields.
- `before_save` (unchanged) → now the single place stamping `availability_changed_on`.
- `on_update()` → publishes `agent_availability_updated` when availability changed.

The endpoint then has nothing left to do, so it's deleted and `desk/src/stores/agentStatus.ts` calls `frappe.client.set_value` directly — the idiom already used for this doctype (`Settings/agents.ts` writes `HD Agent.is_active` the same way). The optimistic write, rollback and socket listener in the store are unchanged; the event now originates from the controller, same room and same payload.

### Permissions

Dropping the endpoint also drops its `agent_only` + `get_agent_name()` self-scoping, and the Agent role has blanket write on **all** HD Agent rows with no `has_permission` hook — so one agent could edit another's record. Added a hook to close it:

- an agent may only modify their own record
- reads stay open, so presence dots and assignment pickers still list every agent
- managers are unrestricted, matching the existing UI contract (the Agents settings tab is already gated on `isAdmin || isManager`)

This was reachable over REST before this PR too, so it's a pre-existing hole being closed rather than a regression introduced here.

Also extracted `is_agent_manager()` in `helpdesk/utils.py` and reused it in `agent_manager_only`, instead of a third copy of the role set.

### Telemetry

Added an `agent_availability_updated` capture, following the `capture_update_telemetry_events` shape already in HD Ticket. Living on the controller means it covers every write path rather than just the desk UI.

The capture and the realtime broadcast share one predicate, `availability_changed()`, which excludes inserts: `has_value_changed` returns `True` when there is no previous doc, so without the guard every agent *created* would be counted as a status change. Excluding the broadcast too costs nothing — `liveStatuses` is a keyed overlay read per agent (`AssignTo.vue:304`), never iterated, so an entry for an agent the client has not fetched is read by nothing; a new agent arrives with the list refetch, which already carries `availability`. `develop` never published on insert either: its only publisher was `set_my_availability`, reachable only by an existing agent changing their own status.

## Testing

`bench run-tests --module helpdesk.helpdesk.doctype.hd_agent.test_hd_agent` — 21 pass; `hd_agent_status` — 15 pass.

New coverage:

- a plain agent can save their own record without elevated rights
- the exact `frappe.client.set_value` payload the store now submits
- an agent **cannot** set another agent's availability; a manager can
- reads of another agent still allowed
- a disabled status is rejected; a since-disabled status doesn't block unrelated saves
- `on_update` publishes on change, and does *not* publish when a save leaves availability untouched

Full app suite compared against a pre-PR baseline on a cleared cache: **identical 44-test failure set** (ERPNext-not-installed plus pre-existing tag/ticket failures), +8 net new tests, zero new failures.

Three test-hygiene fixes were needed along the way: rollback is per-class, so `setUp` now resets availability to a known status — the "did this change?" assertions were otherwise silently dependent on alphabetical test order. `frappe.set_user` is reset per class too, so `setUp`/`tearDown` restore `Administrator`, matching every other test module in the app. And the disabled-status tests toggle the shipped `Unavailable` fixture with an `addCleanup` restore rather than creating new statuses, which leaked across runs (`make_agent_status` can't be made idempotent — `test_hd_agent_status` asserts duplicates raise).

https://claude.ai/code/session_01WuJruCwPQFwBA3HyXogjD7

### DIFF
--- a/helpdesk/api/agent.py
+++ b/helpdesk/api/agent.py
@@ -26,4 +26,3 @@ def sent_invites(emails: list[str], send_welcome_mail_to_user: bool = True):
                 "user_image": user.user_image,
             }
         ).insert()
-    return
--- a/helpdesk/api/agent.py
+++ b/helpdesk/api/agent.py
@@ -1,7 +1,6 @@
 import frappe
-from frappe import _
 
-from helpdesk.utils import agent_only, get_agent_name, publish_event
+from helpdesk.utils import agent_only
 
 
 @frappe.whitelist()
@@ -28,30 +27,3 @@ def sent_invites(emails: list[str], send_welcome_mail_to_user: bool = True):
             }
         ).insert()
     return
-
-
-@frappe.whitelist()
-@agent_only
-def set_my_availability(availability: str) -> dict:
-    if not frappe.db.exists("HD Agent Status", {"name": availability, "enable": 1}):
-        frappe.throw(_("Invalid availability"), frappe.ValidationError)
-
-    name = get_agent_name()
-    if not name:
-        frappe.throw(_("No HD Agent record for current user"), frappe.ValidationError)
-
-    changed_on = frappe.utils.now()
-    frappe.db.set_value(
-        "HD Agent",
-        name,
-        {"availability": availability, "availability_changed_on": changed_on},
-    )
-    publish_event(
-        "agent_availability_updated",
-        data={
-            "agent": name,
-            "availability": availability,
-            "availability_changed_on": changed_on,
-        },
-    )
-    return {"availability": availability}
--- a/helpdesk/helpdesk/doctype/hd_agent/hd_agent.py
+++ b/helpdesk/helpdesk/doctype/hd_agent/hd_agent.py
@@ -6,7 +6,7 @@ from frappe import _
 from frappe.model.document import Document
 
 from helpdesk.helpdesk.doctype.hd_agent_status.hd_agent_status import get_active_status
-from helpdesk.utils import is_agent_manager, publish_event
+from helpdesk.utils import capture_event, is_agent_manager, publish_event
 
 
 class HDAgent(Document):
@@ -46,6 +46,7 @@ class HDAgent(Document):
 
     def on_update(self):
         self.publish_availability_update()
+        self.capture_availability_telemetry()
 
     def validate_availability(self):
         """Only an enabled HD Agent Status may be set as availability.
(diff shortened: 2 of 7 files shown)


helpdesk/api/agent.py
```diff
@@ -28,30 +27,3 @@ def sent_invites(emails: list[str], send_welcome_mail_to_user: bool = True):
             }
         ).insert()
     return
```
RitvikSardana: can you remove this return also?

it is unwanted ig, confirm and if possible remove

aerodeval: Confirmed unwanted — removed in c3c1afa.

It was a bare `return` at the end of the function, so it returned `None` either way; no caller behaviour changes. It was also the one line Codecov flagged as uncovered in this file.

helpdesk/helpdesk/doctype/hd_agent/hd_agent.py
```diff
@@ -39,13 +44,77 @@ def before_save(self):
...
+        a new agent must still be validated.
+        """
+        return bool(self.get_doc_before_save()) and self.has_value_changed(
+            "availability"
+        )
+
+    def publish_availability_update(self):
+        """Broadcast an availability change to every connected client."""
+        if not self.availability_changed():
+            return
```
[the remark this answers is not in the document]
aerodeval: Thanks — checked this carefully, but I don't think the suggestion is right here. Three things:

**1. `liveStatuses` is a keyed overlay, not an iterated source.** Its only reader is `AssignTo.vue:304` — `agentStatusStore.liveStatuses[agent.name]` inside `liveAvailability()`, applied to agents already loaded from `agentResource`. `agentOptions` iterates `agentResource.data`, never `liveStatuses`, and the store is never destructured via `storeToRefs` anywhere. So a broadcast for an agent the client has not fetched writes an entry nothing reads — it cannot surface a new agent's presence.

**2. `develop` never published on insert either.** The only publisher on develop is `helpdesk/api/agent.py:50`, inside `set_my_availability`, reachable only by an existing agent changing their own status. An HD Agent insert has never broadcast in this codebase, so the suggestion would add new behaviour rather than restore removed behaviour.

**3. The "until refresh" path is already covered.** A new agent only appears after a list refetch, and that refetch carries the status — `agentResource` fetches `availability` and `availability_changed_on` (`AssignTo.vue:269-276`), with `liveAvailability` falling back as `live?.availability ?? agent.availability`.

I did verify the inverse risk, which would have been the more serious bug — a genuine status change silently not broadcasting, since `load_doc_before_save` can leave `_doc_before_save` as `None` for a non-new doc. Measured on all paths: insert → no publish; `doc.save()` update → publishes; `frappe.client.set_value` (the real frontend path) → publishes; `get_doc_before_save()` is populated on updates. No regression.

Leaving as-is.

PR #3638, opened 2026-08-01
fix(email): give agent replies their own Message-Id so threads hold

aerodeval:
## What's wrong

Every email carries a unique id (`Message-Id`). A reply points back at the previous message's id using an `In-Reply-To` header — that's how Gmail and Outlook keep a conversation together.

Agent replies were going out with no `In-Reply-To` at all. So every reply arrived as a brand new conversation in the customer's mailbox. One ticket with five replies became five separate email threads.

Measured on our own instance over one week, agent replies on HD Tickets:

| | |
|---|---|
| replies sent | 1,911 |
| with a threading header | 159 (8%) |
| carrying an id the next reply could use | ~0 |

> If you try to reproduce this on Gmail you may not see it. Every reply has an identical subject, and Gmail also groups by subject, so it usually still looks like one conversation. Outlook relies on the headers, which is where it was reported.

## Why

Two changes landed 77 minutes apart on 2025-05-23:

- **frappe#32651** made the `In-Reply-To` header come from `Communication.message_id`.
- **#2308** deleted the code that filled `Communication.message_id` in.

Each was correct on its own. Together they left that column always empty, so no reply could ever be the parent of the next one. #2308 was right about what it removed — that code copied the parent's id onto every reply, so all replies in a ticket shared a single id.

It stayed hidden for 14 months because incoming mail still landed on the right ticket (frappe checks the Email Queue for that, not just Communication). Nothing looked broken from the agent side. The damage was only ever visible in the customer's mailbox.

Different from #3215, which fixed the *inbound* direction — matching an arriving email to its ticket. This is the outbound header.

## The fix

- Give every agent reply its own id, and send that same id.
- Save the id only once the mail has actually been handed off. A reply that failed to send keeps no id, so the next reply can never point at an email nobody received.
- Choose the parent from the newest reply that has an id, so a portal message — which sends no email — doesn't break the chain.

Not a revert of #2308: the id is generated fresh per message, never copied from the parent.

### Bulk reply

A single reply that fails cleans itself up — the whole request rolls back. Bulk reply catches the failure and only logs it, so the row would have stayed behind with an id the customer never got, and the next reply would have threaded onto it. Saving the id after the send covers that too.

Two unrelated problems in the same function, fixed while there: attachments were linked to tickets before any permission check ran, and the same ticket sent twice in one request produced duplicate files. It now checks every ticket first, then attaches, then replies.

## Side effect

Sending an explicit id means frappe stops adding its `isnotification` header. That header has one reader in the framework, which sets a flag used only by `notify_unreplied()` — off by default and never enabled by helpdesk. Core already omits it for sent mail. Reversible with one kwarg if it ever matters.

## Tests

Ten added. This path had none, which is how it went unnoticed.

- a ticket raised by email — the first reply answers the customer, later replies follow the one before them
- two replies get different ids, and the second points at the first
- the id we send matches the id we saved
- a portal message in the middle doesn't break the chain
- no id is saved when email is off, when the send failed, or when nothing was queued (three tests)
- a reply that follows a failed one threads onto the last reply that actually sent
- bulk reply to a ticket the agent can't write to writes nothing at all
- bulk reply with the same ticket twice replies once, with no duplicate files

`test_hd_ticket`: **64 of 66**. The two failures are `test_hold_time_resolution_time_with_holiday*`, which depend on today's date and fail the same way without this branch. Each new test was verified failing against the code it guards.

## Not included

`References`, the other threading header, is still never sent — frappe's `add_headers` forces an `X-` prefix, so it can't be injected app-side. `In-Reply-To` alone is a valid chain. Worth measuring against Outlook before adding more.

Still to do: an end-to-end check on a bench with real outgoing mail.

### DIFF
--- a/helpdesk/api/ticket.py
+++ b/helpdesk/api/ticket.py
@@ -8,23 +8,30 @@ from helpdesk.utils import agent_only, is_admin
 @agent_only
 def bulk_reply(ticket_ids: list, message: str, attachments: list | None = None):
 
-    link_attachments_to_tickets(attachments, ticket_ids)
-
     if not ticket_ids:
         return
 
-    ticket_ids = list(set(ticket_ids))  # Remove duplicates
+    # dedupe but keep the order the agent picked. set() orders by hash, which varies
+    # per process, and duplicates would attach the same file to a ticket twice
+    ticket_ids = list(dict.fromkeys(ticket_ids))
 
+    # Check every ticket before writing anything, so one ticket the agent cannot
+    # reply to does not leave the rest of the batch half done
+    tickets = []
     for ticket_id in ticket_ids:
         frappe.has_permission("HD Ticket", "write", doc=ticket_id, throw=True)
-        doc = frappe.get_doc("HD Ticket", ticket_id)
+        tickets.append(frappe.get_doc("HD Ticket", ticket_id))
+
+    link_attachments_to_tickets(attachments, ticket_ids)
+
+    for doc in tickets:
         try:
             doc.reply_via_agent(
                 message, to=doc.raised_by, attachments=attachments or []
             )
         except Exception as e:
             frappe.log_error(
-                title=f"Bulk reply failed for ticket {ticket_id}",
+                title=f"Bulk reply failed for ticket {doc.name}",
                 message=str(e),
             )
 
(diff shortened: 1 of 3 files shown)


aerodeval:
review is based on bulk_reply bug  for handling attachments and managing of failed communications created due to that which was already pre-existing so will be addressed in a separate PR as it is out of scope from threading ( currently threading takes place as usual failed reply's will have no msg_id set, threading won't break as the helper looks for the last threadable communication)

CC: @RitvikSardana

PR #3729, opened 2026-08-28
feat: timeline timestamp format preference

aerodeval:
Timeline timestamps were always relative, so finding when something exactly happened meant hovering each one. Add a per-user preference that switches all agent timeline timestamps between relative time and the exact date & time, with the other format available in the tooltip.

**No new setting is introduced.** The preference is stored in frappe's existing `User.show_absolute_datetime_in_timeline` field, so there is no new DocType, no custom field, no fixture and no patch. The only backend change is two lines returning that field from `helpdesk.api.auth.get_user`.

### How it works

- A new `TimelineTimestamp` component replaces the repeated `Tooltip` + `timeAgo` blocks in `EmailArea`, `CommentBox`, `HistoryBox` and `CallArea`
- It reads one boolean off the auth store, which comes from `User.show_absolute_datetime_in_timeline`
- The toggle lives in Settings → Preferences and rides the page's existing Save button, dirty badge and outside-click guard — the same contract as language and timezone
- Saving only the timestamp preference refreshes the auth store rather than reloading the page, so the timeline updates in place

## Default (relative)

timeline relative

## The preference

Settings → Preferences → Language & Time:

preference off

It rides the page's existing dirty state and Save button:

preference on, unsaved

## Exact

Saving updates the timeline in place, without a page reload:

timeline exact

The format not shown stays available in the tooltip:

tooltip shows relative

### Notes for reviewers

- **Permissions:** an Agent has no write permission on User via roles, but frappe's `User.share_with_self()` creates a self-DocShare with `write=1` on every User record, which is what makes this work. Verified against a production site.
- **Precedence:** frappe also has this field on System Settings, and desk resolves it as `user field || system default`. This reads the User row alone, deliberately — an agent who wants relative time keeps it whatever the admin sets. The trade-off is that desk and helpdesk will disagree while that site-wide toggle is on.
- **Customer portal:** `TicketCommunication.vue` stays hardcoded to relative. The portal has no preferences UI, so a customer could never change the format.
- **Behaviour change:** `CallArea` previously wrapped its relative time in `__()`; the shared component does not, matching the other three call sites.
- Requires frappe >= 16, which `pyproject.toml` already pins.


RitvikSardana:
No need to make any changes to timeline, we are working on a new one,

and timeline is always relative, like this happened 2 weeks ago, 2 days ago, 2 hours ago or something

michal-bee:
> No need to make any changes to timeline, we are working on a new one,
> 
> and timeline is always relative, like this happened 2 weeks ago, 2 days ago, 2 hours ago or something

@RitvikSardana  This should be IMHO based on system wide settings, as there is already this one. The same principle applies for Theme settings - doesn't make any sense to me to have Light for CRM and Dark for HelpDesk. 

image

RitvikSardana:
@michal-bee yes makes sense, we will use this setting only, closing this PR.

We will be migrating to the new timeline component.

